PR #6, opened 2020-12-09
cli: add vocabularies import 

FlorianCassayre:
* fixes:
  * `docker-services-cli` command
  * restore `requirements-devel.txt`
  * the dumper not loading the vocabulary type name
  * code style
* move import logic to this module, create licenses data
  * command: `invenio vocabularies import [languages, licenses]`


invenio_vocabularies/records/dumper_extensions.py
```diff
@@ -27,4 +27,8 @@ def dump(self, record, data):
 
     def load(self, data, record_cls):
         """Load data."""
-        data.pop("vocabulary_type", None)
+        vocabulary_type_id = data.get("vocabulary_type_id")
```
lnielsen, line 30: Minor: This should not be added back. I removed it because the dump/load has to produce the exact same record:

```python
rec = Record.get_record(...)
assert rec == Records.loads(rec.dumps())
```

Thus if you add a new property in the dump, it has to be removed again in the load method. This is to ensure that we don't have different record representations when we're getting a record from database or elasticsearch.

FlorianCassayre: Thanks for the explanation, I will fix it differently.

invenio_vocabularies/resources/serializers/schema.py
```diff
@@ -28,7 +28,8 @@ class PresentationVocabularySchema(Schema):
     """Vocabulary presentation."""
 
     id = fields.Str(attribute="id")
-    type = fields.Str(attribute="vocabulary_type")
+    # TODO change for resolved type name
+    type = fields.Int(attribute="vocabulary_type_id")
```
kprzerwa, line 32: @lnielsen is this acceptable for now ? I didn't want to leave it as `None`

tests/records/test_api.py
```diff
@@ -44,9 +41,16 @@ def test_vocabulary_type(app, db):
     db.session.add(vocabulary_type)
     db.session.commit()
     record = Vocabulary.create(
-        {}, metadata={"title": "test-item", "type": "test-type"}
+        {}, metadata={"title": {"en": "test-item"}, "type": "test-type"}
```
FlorianCassayre, line 44: If I'm not mistaken `type` is obsolete too.

invenio_vocabularies/cli.py
```diff
@@ -8,76 +8,122 @@
...
+            value = row[attribute]
+            if attribute in i18n:
+                metadata[attribute][default_language] = value
+            elif any(map(lambda s: value.startswith(s + "_"), i18n)):
+                [prefix_attr, language] = attribute.split("_", 1)
+                metadata[prefix_attr][language] = value
+            elif attribute in other:
+                metadata[attribute] = value
+            else:
+                metadata["props"][attribute] = value
```
slint, line 74: Minor: I think the fact that the `id` key (e.g. for languages that would be `"en"`, `"fr"`, etc.) ends up in `props` instead of the top-level... But I guess there's a bigger discussion on how to manage the "user-friendly" identifier of the vocabularies inside the record metadata (or PID)

slint (approved):
LGTM! One minor concern regarding the `id` import field, but we should discuss probably more on how to handle it

invenio_vocabularies/cli.py
FlorianCassayre: It's intended that the `id` ends up in `props` (maybe should have named it `value` for distinction).

PR #15, opened 2021-01-06
api: uncomment vocabulary backend api to allow facets dereference

ppanero:


invenio_vocabularies/api.py
```diff
@@ -9,82 +9,82 @@
...
+    def get_all(self):
+        """Returns all the vocabulary of this type."""
+        return self.backend.get_all(self.vocabulary_type)
 
-#     def dump_all(self):
-#         """Dumps all the vocabulary of this type."""
-#         return list(map(lambda v: v.dump(), self.get_all()))
+    def dump_all(self):
+        """Dumps all the vocabulary of this type."""
+        return list(map(lambda v: v.dump(), self.get_all()))
```
fenekku, line 81: minor: Huh, I didn't know about `map`. Thanks for exposing that... I still prefer list comprehensions: 

```suggestion
        return [v.dump() for v in self.get_all()]
```

fenekku (approved):
looks fine (I could be really convinced with tests ;) )

invenio_vocabularies/api.py
ppanero: This code will be changed during the next month RDM sprint. It was not clear that it was needed before, and now I am not convinced it is the correct way of doing it. I "just" uncommented existing code. Wouldn't spend time on perfecting something that might disappear in a week time 😓 WDYT? Same for tests

PR #28, opened 2021-03-17
bump invenio-db (formerly: add service.rebuild_index)

max-moser:
closes inveniosoftware/invenio-rdm-records#441

add a new service method for reindexing vocabularies

EDIT: this one basically reverted to a bump of invenio-db, because `rebuild_index` has been moved to `invenio-records-resources` (inveniosoftware/invenio-records-resources/pull/211)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ tests_require = [
 
 # Should follow inveniosoftware/invenio versions
 invenio_search_version = ">=1.4.1,<2.0.0"
-invenio_db_version = ">=1.0.5,<2.0.0"
+invenio_db_version = ">=1.0.8,<2.0.0"
 
 extras_require = {
     "docs": [


invenio_vocabularies/services/service.py
```diff
@@ -94,3 +94,21 @@ def search_request(self, identity, params, record_cls, **kwargs):
             params['_type'] = vocabulary_type
         return super().search_request(
             identity, params, record_cls, **kwargs)
+
+    def rebuild_index(self, identity):
```
max-moser, line 98: i think this method would be a good candidate for adding to the `invenio-records-resources` records service interface?

max-moser: and we *could* then introduce a new permission `can_reindex = [SystemProcess()]` that gets checked there... although that probably adds more complexity than benefits, as this functionality isn't intended to be exposed to the web.
what do you think?

lnielsen: Fully agree. Will you move the method to records-resources?

max-moser: yes, i'll create a PR and link it to the same issue

PR #35, opened 2021-05-11
records: cache pid type to reduce queries

lnielsen:
closes inveniosoftware/invenio-app-rdm#833

### DIFF
--- a/invenio_vocabularies/version.py
+++ b/invenio_vocabularies/version.py
@@ -13,4 +13,4 @@ This file is imported by ``invenio_vocabularies.__init__``,
 and parsed by ``setup.py``.
 """
 
-__version__ = "0.6.0"
+__version__ = "0.6.1"


invenio_vocabularies/records/systemfields/pid.py
```diff
@@ -93,10 +89,26 @@ def resolve(self, pid_value):
...
-        return getattr(self, '_type_id', None)
+        # This ensures that when we use Vocabulary.pid.with_type_ctx('...')
+        # we cache the pid type to avoid querying the database every time.
+        type_id = getattr(self, '_type_id', None)
+        if type_id:
+            pid_type = getattr(self, '_pid_type', None)
+            if pid_type is None:
+                pid_type = self.get_pid_type(type_id)
+                self._pid_type = pid_type
+            return pid_type
```
fenekku, line 111: I don't know if this would just add more complexity, but we are starting to do a lot of caching everywhere, so I am wondering if a pattern can emerge from this that would reduce the boilerplate caching code / make the interface nicer (my vote is on not relying on `getattr` too much). Python 3.8 has `cached_property` similar to werkzeug's one, but something along those lines could be implemented (so available in 3.7).

On a bigger level, a lot of this has to do with SQLAlchemy... I am sure this is what you ended up having to do and I am trying to think of a way to make this easier to approach... At this point we have a large data layer that sits on top of SQLAlchemy. Would it be too "out there" to never cache SQLAlchemy results directly, but rather the essential representations that can be loaded/dumped. This is essentially what we are doing here, but expand it elsewhere. It feels like what we we want more and more is:

- perform explicit data access to DB
- manipulate representation but don't save to storage yet
- do further access using representation and searching DB only explicitly - most access is through pre-loaded representations
- then saving to storage
- doing more manipulation / access and so on

i.e. mostly manipulate in memory representations and only access storage *explicitly* - in other words exposing the ORM (session) directly has basically become the exception rather than the norm. The pattern `record.commit()` is a case in point for me: only now sync object with storage ... Strangely this is basically moving us more towards an ActiveRecord / Django ORM data pattern. 

Not sure I am making a concrete point here as much as thinking out loud about the mental model should be for how to deal with our data. :smile:

fenekku (approved):

invenio_vocabularies/records/systemfields/pid.py
```diff
@@ -93,10 +89,26 @@ def resolve(self, pid_value):
...
+        return vocab_type.pid_type
+
     @property
-    def type_id(self):
+    def pid_type(self):
         """Get the current defined type."""
-        return getattr(self, '_type_id', None)
+        # This ensures that when we use Vocabulary.pid.with_type_ctx('...')
+        # we cache the pid type to avoid querying the database every time.
+        type_id = getattr(self, '_type_id', None)
```
ppanero, line 105: where would `_type_id` be set? I cannot find it so Im thinking it would always be None? in most cases it was changes by `_pid_type`, which is set below in line 110...

invenio_vocabularies/records/systemfields/pid.py
ppanero: I agree with this, but I am not sure I understand what the proposal is? 😅

Not sure if we can avoid it tho. Having things in memory allows us to do less queries and does nto have a bit impact on memory after all. Hopefully invalidation does not hit us.

Issue #36, opened 2021-05-18
ui faceting via vocabulary label


ppanero:
The facet class should created in `/services/facets.py` and called `VocabularyLabels`. It will be used from the service configuration, e.g. [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L44).

This class will be set as `value_labels` attribute of e.g. the `TermsFacet` or any other inheriting from `LabelledFacetMixin`. Therefore, it will be called when [`get_label_mapping` is invoked](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/facets/facets.py#L29).

The label class should implement the `__call__` method and return a dict of {`id`: `label`}. See for example the [`RecordRelationLabel`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/facets/labels.py#L14).

Questions:

- How do we deal with i18n? i.e. how do we know which key inside the props (`en`, `de`, etc.) do we use?
- Is it worth implementing an "interface" to be clear which methods should be implemented by `XYZLabels` classes? (i.e. `__call__`), so the way of calling in the `XYZFacets` classes is always the same? On the other hand, if `__call__` is not implemented it will throw an exception so we have an "implicit" interface.

EDIT, discussion IRL:

- How to access the service: for generic vocabularies we will use the available proxy, for those that require specific details and services there will be another labelling class that will need to be aware of the service (e.g. recieve it in the constructor)
- How to access the identity: it is required by the service, but it should suffice with the `AnonymousIdentity` since there are no permissions enforeced there (they are but the policy is `AnyUser`).
- How to deal with i18n: make it into a lazy function, in a similar way than `lazy_gettex` (use `speaklater`) and let the labelling system get it later on. Use [Marshmallow-Utils:gettext_from_dict](https://github.com/inveniosoftware/marshmallow-utils/blob/master/marshmallow_utils/fields/babel.py#L191) as function.

PR #37, opened 2021-05-19
services: add facet labelling

ppanero:
closes https://github.com/inveniosoftware/invenio-vocabularies/issues/36

### DIFF
--- a/invenio_vocabularies/services/facets.py
+++ b/invenio_vocabularies/services/facets.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary facets."""
+
+from flask_principal import AnonymousIdentity
+from invenio_i18n.ext import current_i18n
+from marshmallow_utils.fields.babel import gettext_from_dict
+from speaklater import make_lazy_string
+
+from ..proxies import current_service
+
+
+def lazy_get_label(vocab_item):
+    """Lazy evaluation of a localized vocabulary label."""
+    params = {
+        "locale": current_i18n.locale,
+        "default_locale": "en"
+    }
+
+    return make_lazy_string(
+        gettext_from_dict, vocab_item, **params)
+
+
+class VocabularyLabels:
+    """Fetching of vocabulary labels for facets."""
+
+    def __init__(self, vocabulary, cache=False):
+        """Initialize the labels."""
+        self.vocabulary = vocabulary
+        self.cache = cache
+        self.fields = ["id", "title"]  # not configurable
+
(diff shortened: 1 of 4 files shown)


tests/services/conftest.py
```diff
@@ -16,15 +16,6 @@
 from invenio_vocabularies.records.api import Vocabulary
 
 
-@pytest.fixture()
```
ppanero, line 19: it was repeated with base `conftest.py`, and needed to be of scope `module` to work with `example_records`

setup.py
```diff
@@ -60,7 +60,7 @@
 
 install_requires = [
     "flask>=1.1,<2",
-    "invenio-records-resources>=0.15.1,<0.16.0",
```
ppanero, line 63: needed to get `read_all` and `read_many` methods

invenio_vocabularies/services/facets.py
```diff
@@ -0,0 +1,63 @@
...
+            # cannot loop over ids because vocab id is inside each item
+            if len(ids) == len(seen):
+                break
+
+            id_ = vocab["id"]
+            if id_ in ids:
+                labels[id_] = lazy_get_label(vocab["title"])
+                seen.add(id_)
+
+        return labels
```
fenekku, line 63: This seemed complicated, so hopefully I didn't *oversimplify*. 

```suggestion
        return {
            vocab["id"]: lazy_get_label(vocab["title"])
            for vocab in vocab.hits if vocab["id"] in set(ids)
        }
```

Previous code seemed to imply there could be more than 1 result with same "id" in `vocab.hits`:
- two vocab hits with same "id" have same other values, right?
- `read_all` doesn't return duplicates, right?
-  Only `read_many` could and that's because of line 39 `ids` argument that may contain duplicate, right?
- So if we really want to prevent that (not really sure if it makes a significant difference at this point :smile: ), we could `set(ids)` before passing them to `read_many`.

fenekku (approved):
sidenote: We might need a variation on this for resource types :+1:

invenio_vocabularies/services/facets.py
ppanero: I think it was not fully understood, maybe I need to put more comments:

- In the suggested code (with the `set(ids)` inside the for loop, you would be casting to set on *every* iteration, which in the long run might be costly.
- The idea behind having the `seen` set was not about duplicates (which cannot happen, ever unless something doggy happens). The point was to have some sort of "optimistic" optimization where there is no need to loop over the whooooole vocbulary, and stop when all ids have been seen.

Imagine the following case:

```
vocab.hits = [1, 2, 3, 4, 5 .... 1000000, 1000001]
ids = [1, 2, 3]
```

With the code as it was, it would stop and return in the 3rd iteration. With the suggested one it would loop 1000001 times. Obviously if the used asks for id 1000001 there is nothing we can do...

ppanero: Discuss IRL, leaving as is for now, we can revist in the future.

PR #40, opened 2021-05-26
services: add support for filtered read_all

### LINKED ISSUE #38: Override read_all to filter by type like search
invenio-vocabularies's overrides its search method to filter by type:

def search(self, identity, params=None, es_preference=None, type=None,  # <-- this
           **kwargs)

. A similar `read_all` is needed in `VocabulariesService` to only read all a specific vocabulary type documents. 

In particular, this is needed if we want to use `read_all` to get all possible resource types in the deposit page dropdown.

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/250
closes #38 

- Note there is a parameter order change in the service methods to make the vocabulary type mandatory
- No tests are added since the method was already tested by the `test_facets.py` (not a valid excuse, but is good for now)

### DIFF
--- a/invenio_vocabularies/services/service.py
+++ b/invenio_vocabularies/services/service.py
@@ -9,7 +9,9 @@
 """Vocabulary service."""
 
 from elasticsearch_dsl.query import Q
+from elasticsearch_dsl.response import Response
 from flask_babelex import lazy_gettext as _
+from invenio_cache import current_cache
 from invenio_db import db
 from invenio_records_resources.services import Link, LinksTemplate, \
     RecordService, RecordServiceConfig, SearchOptions, pagination_links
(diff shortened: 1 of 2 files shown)


ppanero:
Once merged needs to be migrated here https://github.com/inveniosoftware/invenio-app-rdm/pull/853/files#r638535062 too

invenio_vocabularies/services/service.py
```diff
@@ -105,7 +107,7 @@ def create_type(self, identity, id, pid_type):
         db.session.commit()
         return type_
 
-    def search(self, identity, params=None, es_preference=None, type=None,
+    def search(self, identity, type, params=None, es_preference=None,
```
lnielsen, line 110: This will break with this change https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/resources/resource.py#L106

Now the question is why the tests don't break 😄 

To avoid problems before release, please revert this change

invenio_vocabularies/services/service.py
```diff
@@ -135,3 +137,35 @@ def search(self, identity, params=None, es_preference=None, type=None,
...
         )
+
+    def read_all(self, identity, fields, type, cache=True, **kwargs):
+        """Search for records matching the querystring."""
+        cache_key = "-".join(fields)
+        results = current_cache.get(cache_key)
+        es_query = Q("match_all")
+
+        # If not found, NoResultFound is raised (caught by the resource).
+        vocabulary_type = VocabularyType.query.filter_by(id=type).one()
```
lnielsen, line 148: This will perform a database query and should be avoided when the results are cached.

invenio_vocabularies/services/service.py
ppanero: Tests do not break because the call is made with "kwargs" i.e. type=... not just assuming the "position". But yes, ill revert.

invenio_vocabularies/services/service.py
ppanero: we wont be able to reconstruct the search fully then... but since there are no links it should be potentially fine...

PR #42, opened 2021-06-03
facets: add nested vocabulary facet and label

ppanero:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/628

Example output for resource_types:

```
{
  "aggregations": {
    "resource_type": {
      "buckets": [
        {
          "key": "image",
          "doc_count": 10,
          "label": "Image",
          "is_selected": false,
          "inner": {
            "buckets": [
            {
              "key": "image-diagram",
              "doc_count": 4,
              "label": "Diagram",
              "is_selected": false
            },
            {
              "key": "image-other",
              "doc_count": 6,
              "label": "Other",
              "is_selected": false
            }
          ]
        },
        ...
```

Screenshot 2021-06-07 at 08 55 58

### DIFF
--- a/invenio_vocabularies/services/facets/facets.py
+++ b/invenio_vocabularies/services/facets/facets.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary facets."""
+
+from bisect import bisect, bisect_left
+
+from elasticsearch_dsl import AttrDict
+from invenio_records_resources.services.records.facets import TermsFacet
+
+
+class LabellingBucket(AttrDict):
+    """Bucket class to allow label extration."""
+
+    def __init__(self, key, doc_count):
+        """Constructor."""
+        super().__init__({"key": key, "doc_count": doc_count})
+
+
+class NestedVocabularyTermsFacet(TermsFacet):
+    """A term facet for vocabularies with nested/hierarchical labelling."""
+
+    def __init__(self, splitchar="-", **kwargs):
+        """Initialize the nested vocabulary terms facet."""
+        self._splitchar = splitchar
+        super().__init__(**kwargs)
+
+    def _build_nested_buckets(self, buckets):
+        """Build nested buckets from a vocabulary facet.
+
+        This function will return a flattened list of buckets. The input
+        will be a list of ES buckets with, potentially, a nested key.
+        That means, dash separated values (e.g. image-drawing). The output
+        be [image, image-drawing], so the parent (i.e. image) of the nested
+        aggregation is also queried to ES (to get the UI label).
+        In addition, it will calculate the aggregated value of the parent
+        doc_count.
+
+        For example, doc_count in parenthesis:
+        bucket.data = [
+            image-png(2), image-jpg(5), publication(3), publication-book(1)
+        ]
+        buckets_out = [
+            image(7), image-jpg(5), image-png(2),
+            publication(4), publication-book(1)
+        ]
+        """
+        seen = set()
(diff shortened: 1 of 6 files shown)


invenio_vocabularies/services/facets/facets.py
```diff
@@ -0,0 +1,127 @@
...
+        bucket_out = {
+            "key": key,
+            "doc_count": self.get_metric(bucket),
+            "label": label_map[key],
+            "is_selected": self.is_filtered(full_key, filter_values)
+        }
+
+        return bucket_out
+
+    def _build_bucket_list(
```
ppanero, line 76: Not a fan of how this function looks... the logic is more than complex... need to rethink it

invenio_vocabularies/services/facets/facets.py
```diff
@@ -0,0 +1,129 @@
...
+    """A term facet for vocabularies with nested/hierarchical labelling."""
+
+    def __init__(self, splitchar="-", **kwargs):
+        """Initialize the nested vocabulary terms facet."""
+        self._splitchar = splitchar
+        super().__init__(**kwargs)
+
+    def _build_nested_buckets(self, buckets):
+        """Build nested buckets from a vocabulary facet."""
+        seen = set()  # cheaper insert and in operations
```
ntarocco, line 35:
```suggestion
        seen = set()
```

invenio_vocabularies/services/facets/facets.py
```diff
@@ -0,0 +1,129 @@
...
+                    idx = bisect_left(buckets_idx, root)
+                    buckets_out[idx].doc_count += metric
+
+            if key not in seen:  # Add the last bucket
+                seen.add(key)
+                idx = bisect(buckets_idx, key)
+                buckets_out.insert(idx, bucket)
+                buckets_idx.insert(idx, key)  # add full key to idx
+
+        return buckets_out
```
ntarocco, line 61: I am sure there are a lot of smart things happening here, but it is kind of hard to figure out what this list will contain at the end, in particular with nested buckets.

For these given buckets:
* image
* image-png
* image-jpg

Is it returning `[image, image-png, image-jpg]`?

Other question: are the bucket returned by ES alphabetically sorted? Or sort by metric descending? Do we have to re-sort them, for example to have instead: `[image, image-jpg, image-png]`
This is what I would try to add to the method docstring to understand better what is happening here.

tests/services/test_facets.py
```diff
@@ -10,68 +10,152 @@
...
-
-def test_vocabulary_label_not_found(example_records):
-    # Uses example record to have two languages created on the vocabulary
-    labels_instance = VocabularyLabels(vocabulary="languages", cache=True)
-    labels = labels_instance(["eng", "glg"])
-
-    assert labels["eng"] == "English"
-    assert not labels.get("glg")  # search wont fail
+@pytest.fixture(scope="module")
+def nested_item_labels(app, nested_vocabulary):
```
ntarocco, line 111: Just IMO: I would have added this to the test and not as a fixture if not re-usable.

invenio_vocabularies/services/facets/facets.py
ppanero: The comment is correct. Imagine this two cases:

```
buckets.data = [ image-png, image-jpg, publication]
buckets_out = [image, image-png, image-jpg, publication]
```

and

```
buckets.data = [image, image-png, image-jpg, publication]
buckets_out = [image, image-png, image-jpg, publication]
```

As for the order, they do not... that's why I am using the bisect library to be able to have a sorted `bucket_out` and need to maintain a sorted `bucket_idx` with just the key (not the buckets). Otherwise, I would need to re-implement the `__lt__` function so bisect can properly find the insertion point.

tests/services/test_facets.py
ppanero: That's true, it just seemed cleaner with the values in fixtures. But changed :)

ntarocco:
I would propose to leave it as it, and eventually refactor the readability of the functions if they will create misunderstandings in the future.

ntarocco (approved):

PR #47, opened 2021-06-15
services: read many implemented and tested

### LINKED ISSUE #43: Implement read_many
When serializing to datacite, having a read_many to retrieve multiple subjects (vocabularies) at once would be great.

### LINKED ISSUE #46: tests: read_all and read_many for vocabularies
test methods before PR
Closes #43

effervescent-shot:
Closes #43 
Closes #46

### DIFF
--- a/invenio_vocabularies/services/service.py
+++ b/invenio_vocabularies/services/service.py
@@ -168,3 +168,20 @@ class VocabulariesService(RecordService):
             results = Response(search, results)
 
         return self.result_list(self, identity, results)
+
+    def read_many(self, identity, type, ids, fields=None, **kwargs):
+        """Search for records matching the querystring filtered by ids."""
+        es_query = Q("match_all")
+        vocabulary_type = VocabularyType.query.filter_by(id=type).one()
+        vocab_id_filter = Q('term', type__id=vocabulary_type.id)
+        filters = []
+
+        for id_ in ids:
+            filters.append(Q('term', **{"id": id_}))
+        filter = Q('bool', minimum_should_match=1, should=filters)
+        filter = filter & vocab_id_filter
+        results = self._read_many(
+            identity, es_query, fields,
+            extra_filter=filter, **kwargs)
+
+        return self.result_list(self, identity, results)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -96,7 +96,7 @@ def lang_type(db):
     return VocabularyType.create(id='languages', pid_type='lng')
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='function')
 def lang_data():
     """Example data."""
     return {
@@ -135,10 +135,23 @@ def example_record(db, identity, service, example_data):
     record = service.create(
         identity=identity,
         data=dict(
-            **example_data, vocabulary_type_id=vocabulary_type_languages.id
+            **example_data,
+            vocabulary_type_id=vocabulary_type_languages.id
         ),
     )
 
     Vocabulary.index.refresh()  # Refresh the index
-
     return record
+
+
+@pytest.fixture(scope='function')
+def lang_data_many(lang_type, lic_type, lang_data, service, identity):
+    """Create many language vocabulary."""
+    lang_ids = ['fr', 'tr', 'gr', 'ger', 'es']
+    data = dict(lang_data)
+
+    for lang_id in lang_ids:
+        data['id'] = lang_id
+        service.create(identity, data)
+    Vocabulary.index.refresh()  # Refresh the index
+    return lang_ids
--- a/tests/services/test_service.py
+++ b/tests/services/test_service.py
@@ -136,3 +136,18 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
         lang_data2['props']['newkey'] = v
         pytest.raises(
             ValidationError, service.create, identity, lang_data2)
+
+
+def test_read_all_no_cache(lang_type, lang_data_many, service, identity):
+    """ read_all method should return all languages created in this scope. """
+    items = service.read_all(identity, fields=[],
+                             type='languages', cache=False)
+    assert set(lang_data_many).issubset(set([i.id for i in items._results]))
+
+
+def test_read_many(lang_type, lang_data_many, service, identity):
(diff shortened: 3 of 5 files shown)


invenio_vocabularies/services/service.py
```diff
@@ -139,7 +139,8 @@ def search(self, identity, params=None, es_preference=None, type=None,
         )
 
     def read_all(self, identity, fields, type, cache=True, **kwargs):
-        """Search for records matching the querystring."""
+        """Search for records matching the querystring.""" 
+
```
ppanero, line 143: pycodestyle: no new lines after docstrings :P

invenio_vocabularies/services/service.py
```diff
@@ -168,3 +169,24 @@ def read_all(self, identity, fields, type, cache=True, **kwargs):
             results = Response(search, results)
 
         return self.result_list(self, identity, results)
+
+
+    def read_many(self, identity, type, ids, fields=None, **kwargs):
+        """Search for records matching the querystring.""" #TODO: Write a better doc string
+
```
ppanero, line 176: pycodestyle: no new lines after docstrings :P

invenio_vocabularies/services/service.py
```diff
@@ -168,3 +169,24 @@ def read_all(self, identity, fields, type, cache=True, **kwargs):
...
+        es_query = Q("match_all")
+        vocabulary_type = VocabularyType.query.filter_by(id=type).one() 
+        vocab_id_filter = Q('term', type__id=vocabulary_type.id)
+        
+        filters = []
+        for id_ in ids:
+            filters.append(Q('term', **{"id": id_}))
+            
+        query = Q('bool', minimum_should_match=1, should=filters)
```
ppanero, line 185: minor: since this is passed as `extra_filter` I would call it `filter`

tests/conftest.py
```diff
@@ -142,3 +143,22 @@ def example_record(db, identity, service, example_data):
...
+
+
+@pytest.fixture(scope='function')
+def lang_data_many(lang_type, lang_data, service, identity):
+    """Create many language vocabulary."""
+    
+    lang_ids = ['fr','tr','gr','ger','es']
+    data = deepcopy(dict(lang_data))
+
+    if 'props' in data.keys():
```
ppanero, line 155: If `props` are not needed/supported, I think we should remove from the source fixture

tests/services/test_service.py
```diff
@@ -136,3 +136,22 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
...
+    #assert len(items._results) == len(lang_data_many) #TODO: Fix it to actual number with scope='module'
+
+
+def test_read_many(lang_type, lic_type, lang_data_many, service, identity):
+    """ read_many method should return all requested languages. """
+
+    items = service.read_many(identity, type='languages', ids=['fr','tr','es'], fields=[])
+    item_ids = [i.id for i in items._results]
+    for id_ in ['fr','tr','es']:
+        assert id_ in item_ids
```
ppanero, line 155: faster with sets?
```python
assert {'fr', 'tr', 'es'} == set(item_ids)
```

Shouldn't be a subset, since they should be the *only* ones returned.

tests/services/test_service.py
```diff
@@ -136,3 +136,22 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
...
         pytest.raises(
             ValidationError, service.create, identity, lang_data2)
+
+
+def test_read_all_no_cache(lang_type, lic_type, lang_data_many, service, identity):
+    """ read_all method should return all languages created in this scope. """
+
+    items = service.read_all(identity, fields=[], type='languages', cache=False)
+    assert set(lang_data_many).issubset(set([i.id for i in items._results])) 
+    #assert len(items._results) == len(lang_data_many) #TODO: Fix it to actual number with scope='module'
```
ppanero, line 146: wouldn't it better to keep it scope to this function... no need to check the length as long as the known existing ones are there.

what about actually check that the vocabulary *type* (i.e. languages) is correct in all of them, and no other vocabulary type got in the results.

tests/services/test_service.py
```diff
@@ -136,3 +136,22 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
         lang_data2['props']['newkey'] = v
         pytest.raises(
             ValidationError, service.create, identity, lang_data2)
+
+
+def test_read_all_no_cache(lang_type, lic_type, lang_data_many, service, identity):
```
ppanero, line 141: Since cache/no cache is checked at records-resources level... I think is safe to just tests that we only get vocabularies of one type.
```suggestion
def test_read_all(lang_type, lic_type, lang_data_many, service, identity):
```

ppanero (commented):
Nice work! 🚀  Functionality is GTG, just some minor cosmetic issues and a few test enhancements.

tests/services/test_service.py
effervescent-shot: elegant, thought of it after push

tests/services/test_service.py
effervescent-shot: still, we can write a basic test if it works for vocab so I left it

ppanero: Sounds good :)

invenio_vocabularies/services/service.py
ppanero: This has been resolved but nor changed nor answered 😅 WDYT about the possible name change?

effervescent-shot: done

tests/services/test_service.py
```diff
@@ -136,3 +136,18 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
         lang_data2['props']['newkey'] = v
         pytest.raises(
             ValidationError, service.create, identity, lang_data2)
+
+
+def test_read_all_no_cache(lang_type, lang_data_many, service, identity):
```
ppanero, line 141: For future reference, maybe to be removed in a next PR. If `lang_data_many` uses `lang_type` there is no need to apply it here :)

ppanero:
Merging... lowest takes ages to resolve deps.

PR #49, opened 2021-06-16
jsonschemas: redefine using definitions

### LINKED ISSUE #48: records: refactor jsonschema to use definitions
- Refactor vocabularies schema to use JSONRef definitions. Copy from RDM-Records ``records/definitions.json`` ``internal-pid`` schema to invenio-records-resources. 
    - Add def for ``$schema``, ``id`` in invenio-records-resources.

- Create a ``invenio_vocabularies/records/jsonschemas/vocabularies/definitions-v1.0.0.json``
    - Add ``title``, ``description``, ``icon``.

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/262
closes #48 

**open questions**
> The initial model contains `name` and no `descriptions`. However, the issue mentions it. Should, `name` be `description` and therefore use the definition? The main difference would be localization (i.e. the first has not, the second does)

- `name` (string)
- `acronym` (list) should be there
- `descriptions` should be removed
- `titles` stays for search purposes but there is no localization, UI will use the `name`
- `identifiers` should allow a list, but we only take a `RoR` from the `dump` (use `uniqueItems`). Set allowing multiple `RoR` but not twice the same `RoR` value.

> There are no `jsonschemas` in invenio-records-resources, but it might make sense to have basic fields (`$schemas`, `id`) as a definition. I'm checking if we can resolve from other modules. Did not find it possible with the `.refresolver_store()` since the `schemas` seem to be extension binded. 

Discussed IRL, move to `jsonschemas/definitions-v1.0.0.json` to `invenio-records-resources` (then change also `rdm-records` to use said definitions)


invenio_vocabularies/contrib/affiliations/jsonschemas/affiliations/affilation-v1.0.0.json
```diff
@@ -5,49 +5,28 @@
   "additionalProperties": false,
```
fenekku, line 5: there is a typo in the name of the file, it should be:

`affilation-v1.0.0.json` -> `affiliation-v1.0.0.json`

invenio_vocabularies/records/jsonschemas/vocabularies/definitions-v1.0.0.json
```diff
@@ -0,0 +1,14 @@
+{
+  "title": {
+    "type": "object",
+    "description": "Title of vocabulary item. Key are locale codes.",
```
fenekku, line 4: minor
```suggestion
    "description": "Title of vocabulary item. Keys are locale codes.",
```

invenio_vocabularies/records/jsonschemas/vocabularies/vocabulary-v1.0.0.json
```diff
@@ -5,72 +5,51 @@
...
+    "props": {
       "type": "object",
-      "description": "Description of vocabulary item. Key are locale codes.",
       "patternProperties": {
-        "^[a-z]{2}$": {
+        "^.*$": {
           "type": "string"
         }
       }
     },
```
fenekku, line 35: We probably don't have to specify the key pattern here (it looks pretty generic anyway :smile: )

fenekku (approved):
Mergeable after comments have been addressed :+1:

invenio_vocabularies/records/jsonschemas/vocabularies/vocabulary-v1.0.0.json
ppanero: hmm I think we want to make is "tight" to localization so it should be 2 and alphabetic (not numeric)... Otherwise we can end up with pretty weird things indexed... No?

ppanero: If we agree on changing this, it can be addressed in another PR :)

ppanero:
Merging... lowerst takes ages to resolve deps

invenio_vocabularies/records/jsonschemas/vocabularies/vocabulary-v1.0.0.json
fenekku: This `patternProperties` is for `props` though and that can be anything I thought.

PR #51, opened 2021-06-17
contrib: affiliations vocabulary

ppanero:
requires https://github.com/inveniosoftware/invenio-vocabularies/pull/49
partly closes https://github.com/inveniosoftware/invenio-rdm-records/issues/328

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/affiliations.py
+++ b/invenio_vocabularies/contrib/affiliations/affiliations.py
@@ -13,16 +13,23 @@ from invenio_records_resources.factories.factory import RecordTypeFactory
 from ...records.pidprovider import PIDProviderFactory, VocabularyIdProvider
 from ...records.systemfields import BaseVocabularyPIDFieldContext
 from ...services.permissions import PermissionPolicy
+from .config import AffiliationsSearchOptions, service_components
 from .schema import AffiliationSchema
 
-affiliation_record_type = RecordTypeFactory(
+record_type = RecordTypeFactory(
     "Affiliation",
-    AffiliationSchema,
-    permission_policy_cls=PermissionPolicy,
-    pid_provider_cls=PIDProviderFactory.create(pid_type='aff'),
-    pid_ctx_cls=BaseVocabularyPIDFieldContext,
-    pid_create=False,
+    # Data layer
+    pid_field_kwargs={
+        "create": False,
+        "provider": PIDProviderFactory.create(pid_type='aff'),
+        "context_cls": BaseVocabularyPIDFieldContext,
+    },
     schema_version="1.0.0",
     schema_path="local://affiliations/affiliation-v1.0.0.json",
-    endpoint_route='/affiliations'
+    # Service layer
+    endpoint_route='/affiliations',
+    service_schema=AffiliationSchema,
+    search_options=None,
+    service_components=service_components,
+    permission_policy_cls=PermissionPolicy,
 )
--- a/invenio_vocabularies/contrib/affiliations/affiliations.py
+++ b/invenio_vocabularies/contrib/affiliations/affiliations.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary affiliations."""
+
+from invenio_records_resources.factories.factory import RecordTypeFactory
+
+from ...records.pidprovider import PIDProviderFactory, VocabularyIdProvider
+from ...records.systemfields import BaseVocabularyPIDFieldContext
+from ...services.permissions import PermissionPolicy
+from .schema import AffiliationSchema
+
+affiliation_record_type = RecordTypeFactory(
+    "Affiliation",
+    AffiliationSchema,
+    permission_policy_cls=PermissionPolicy,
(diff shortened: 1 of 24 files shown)


invenio_vocabularies/records/mappings/v7/vocabularies/vocabulary-v1.0.0.json
```diff
@@ -3,7 +3,8 @@
     "dynamic": "strict",
     "properties": {
       "$schema": {
-        "type": "keyword"
+        "type": "keyword",
+        "index": "false"
```
ppanero, line 7: Assuming we never search by `$schema`

invenio_vocabularies/records/mappings/v7/vocabularies/vocabulary-v1.0.0.json
```diff
@@ -71,7 +72,8 @@
         "dynamic": true
       },
       "icon": {
-        "type": "keyword"
+        "type": "keyword",
+        "index": "false"
```
ppanero, line 76: Assuming we never search by `icon`

invenio_vocabularies/records/pidprovider.py
```diff
@@ -59,3 +59,61 @@ def create(cls, object_type=None, object_uuid=None, record=None, **kwargs):
             object_uuid=object_uuid,
             status=PIDStatus.REGISTERED
         )
+
+
+class CustomVocabularyPIDProvider(BaseProvider):
```
ppanero, line 64: need help: naming? 🤔 If we call this *Custom*, should we call the other one *Generic* i.e.

```diff
- VocabularyIdProvider
+ GenericVocabularyPIDProvider
```

invenio_vocabularies/contrib/affiliations/affiliations.py
```diff
@@ -0,0 +1,33 @@
...
+
+from invenio_records_resources.factories.factory import RecordTypeFactory
+
+from ...records.pidprovider import PIDProviderFactory, VocabularyIdProvider
+from ...records.systemfields import BaseVocabularyPIDFieldContext
+from ...services.permissions import PermissionPolicy
+from .config import AffiliationsSearchOptions, service_components
+from .schema import AffiliationSchema
+
+affiliation_record_type = RecordTypeFactory(
```
lnielsen, line 19: Shorter as it's already in a module and package both named affiliations.
```suggestion
record_type = RecordTypeFactory(
```

invenio_vocabularies/contrib/affiliations/affiliations.py
```diff
@@ -0,0 +1,33 @@
...
+    "Affiliation",
+    # Data layer
+    pid_provider_cls=PIDProviderFactory.create(pid_type='aff'),
+    pid_ctx_cls=BaseVocabularyPIDFieldContext,
+    pid_create=False,
+    schema_version="1.0.0",
+    schema_path="local://affiliations/affiliation-v1.0.0.json",
+    # Service layer
+    endpoint_route='/affiliations',
+    record_type_service_schema=AffiliationSchema,
```
lnielsen, line 29: Probably we should rename this one Records-Resources to just ``service_schema`` similar to e.g. ``service_components``

invenio_vocabularies/contrib/affiliations/config.py
```diff
@@ -0,0 +1,62 @@
...
+            'id.text._3gram',
+            'title.en^5',
+            'title.en._2gram',
+            'title.en._3gram',
+        ],
+    )
+
+    sort_default = 'bestmatch'
+
+    sort_default_no_query = 'name'
```
lnielsen, line 36: Name is not defined in ``sort_options`` below.

invenio_vocabularies/records/pidprovider.py
lnielsen: Yeah, not easy. In general for naming, it's good to think of what the class does and not for what it's used  - e.g. in this case it takes an id provided by the user so perhaps something like ``UserSpecifiedPIDProvider`` but I don't really like it.

invenio_vocabularies/services/schema.py
```diff
@@ -24,15 +24,15 @@ class BaseVocabularySchema(BaseRecordSchema):
...
-    description = i18n_strings
-    icon = fields.Str(allow_none=False)
 
 
 class VocabularySchema(BaseVocabularySchema):
     """Service schema for vocabulary records.."""
 
-    tags = fields.List(SanitizedUnicode())
-    type = fields.Str(attribute='type.id', required=True)
+    description = i18n_strings
```
lnielsen, line 32: They cannot be left above? Even id they are not used in the affiliations vocab. From the specs we wrote, we chose this as common props.

invenio_vocabularies/contrib/affiliations/config.py
ppanero: True, I had solved that in a PR for resource level since is not fully seen at service level. corrected to `title` for now.

invenio_vocabularies/services/schema.py
ppanero: You mean to *accept them if they exist*? I thought we wanted to enforce a more restricted schema. I can put them back, but will they always be empty? It would affect also jsonschema and ES mappings.

invenio_vocabularies/records/pidprovider.py
ppanero: Yeah... tbh `CustomVocabularyPIDProvider` is the best I can come up with right now... :/ But I'm not convinced.

PR #52, opened 2021-06-18
test: read_all tested with and without cache

### LINKED ISSUE #50: tests: implement tests for read_all with and without cache
They got removed from `records-resources` due to an issue with permissions <https://github.com/inveniosoftware/invenio-records-resources/pull/261>

effervescent-shot:
Closes #50

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,7 @@ import pytest
 from flask_principal import Identity, Need, UserNeed
 from invenio_access.permissions import any_user, system_process
 from invenio_app.factory import create_api as _create_api
+from invenio_cache import current_cache
 
 from invenio_vocabularies.records.api import Vocabulary
 from invenio_vocabularies.records.models import VocabularyType
@@ -155,3 +156,14 @@ def lang_data_many(lang_type, lic_type, lang_data, service, identity):
         service.create(identity, data)
     Vocabulary.index.refresh()  # Refresh the index
     return lang_ids
+
+
+# FIXME: https://github.com/inveniosoftware/pytest-invenio/issues/30
+# Without this, success of test depends on the tests order
+@pytest.fixture()
+def cache():
+    """Empty cache."""
+    try:
+        yield current_cache
+    finally:
+        current_cache.clear()
--- a/tests/services/test_service.py
+++ b/tests/services/test_service.py
@@ -10,6 +10,7 @@
 """Test the vocabulary service."""
 
 import pytest
+from invenio_cache import current_cache
 from invenio_pidstore.errors import PIDAlreadyExists, PIDDeletedError
 from marshmallow.exceptions import ValidationError
 from sqlalchemy.exc import IntegrityError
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -155,3 +156,14 @@ def lang_data_many(lang_type, lic_type, lang_data, service, identity):
         service.create(identity, data)
     Vocabulary.index.refresh()  # Refresh the index
     return lang_ids
+
+
+# FIXME: https://github.com/inveniosoftware/pytest-invenio/issues/30
+# NOTE: Without this, success of test depends on test order
+@pytest.fixture()
+def cache():
```
ppanero, line 164: Nice catch! 🚀 Do you know if we still need the fixture in `invenio-records-resources` or it could also be removed from there?

tests/conftest.py
```diff
@@ -155,3 +156,14 @@ def lang_data_many(lang_type, lic_type, lang_data, service, identity):
         service.create(identity, data)
     Vocabulary.index.refresh()  # Refresh the index
     return lang_ids
+
+
+# FIXME: https://github.com/inveniosoftware/pytest-invenio/issues/30
+# NOTE: Without this, success of test depends on test order
```
ppanero, line 162: I learnt this maybe a few weeks ago... It turns out that `NOTE` has a specific meaning for translations (i.e. this will be picked up by babel)... So for notes we should just put the comment :) If you see any around the code, it was probably me... I'm trying to remove them when I see them 😅 

```suggestion
# without this, the success of test depends on the tests order
```

tests/services/test_service.py
```diff
@@ -138,11 +139,24 @@ def test_missing_or_invalid_type(lang_data2, service, identity):
...
     assert set(lang_data_many).issubset(set([i.id for i in items._results]))
+    cached = current_cache.get("id")
+    assert not cached
+
+
+def test_read_all_cache(lang_type, lang_data_many, service, identity, cache):
+    """ read_all method should return all languages created in this scope. """
+    items = service.read_all(identity, fields=['id'],
+                             type='languages', cache=True)
+    assert set(lang_data_many).issubset(set([i.id for i in items._results]))
```
ppanero, line 157: minor: accessing `_result` is sort of breaking the intention of having it "private" (i.e. with `_`)... I am wondering we should just access `items` and use the dict approach:
```suggestion
    assert set(lang_data_many).issubset(set([i["id"] for i in items]))
```

---

I am also wandering since the `ResultItem` has an `.id` attribute, why does each element of `ResultItemList` not have it... WDYT? (With this I'm trying to trigger a discussion point, maybe there ia another issue to solve). For now, just the changes of what's written above 👼

ppanero (approved):
Nice work! 2 minor comments :) It can be merged once they are addressed.

Maybe we should open an issue for the last discussion point.

tests/conftest.py
effervescent-shot: I'd take a cautious approach. If cache is only required for read_all/read_many then yes it could be removed. I don't have the latest code of invenio-record-resources (I was debugging cache functionality :grin: from the old code)

tests/conftest.py
effervescent-shot: Nice!

tests/services/test_service.py
effervescent-shot: Each item in ResultItemList has an id, otherwise how this code will work?
> assert set(lang_data_many).issubset(set([i["id"] for i in items]))

PR #55, opened 2021-06-21
subtypes: filter suggest + store in db

fenekku:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/899 in part (backend)


fenekku:
The tests were cancelled after dependency resolution took too long, so this should be good to be reviewed + merged.

gen-migration.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
ppanero, line 1: Since this is not a fully automated process - and sometimes doesn't work - I would not commit it here... we can keep a gist somewhere...  It should be something more generic, not just `invenio-vocabularies`.

invenio_vocabularies/alembic/1d3302223885_create_vocabularies_subtype.py
```diff
@@ -0,0 +1,47 @@
...
+# details.
+
+"""create_vocabularies_subtype."""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '1d3302223885'
+down_revision = '4f365fced43f'
```
ppanero, line 16: Not ultra familiar with alembic, I guess it supports branching. i.e. both this and the affiliations recipe have the same `down_version`. Just trying to make sure none will be skipped.

invenio_vocabularies/records/models.py
```diff
@@ -53,3 +53,32 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
...
+class VocabularySubtype(db.Model):
+    """Vocabulary subtype type model.
+
+    Current use case is for subject types.
+    """
+
+    __tablename__ = "vocabularies_subtypes"
+
+    id = db.Column(db.String, primary_key=True)
+    # pid_type = db.Column(db.String, unique=True)
```
ppanero, line 67: remove?

invenio_vocabularies/records/models.py
```diff
@@ -53,3 +53,32 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
...
+
+class VocabularySubtype(db.Model):
+    """Vocabulary subtype type model.
+
+    Current use case is for subject types.
+    """
+
+    __tablename__ = "vocabularies_subtypes"
+
+    id = db.Column(db.String, primary_key=True)
```
ppanero, line 66: I see the vocabulary_type FK is not part of this table PK. What would happen if there are two subtypes with the same `id` (form different type)?

invenio_vocabularies/resources/resource.py
```diff
@@ -52,7 +52,7 @@ class VocabulariesResourceConfig(RecordResourceConfig):
         "type": ma.fields.Str(required=True),
     }
 
-    request_args = VocabularySearchRequestArgsSchema
+    request_search_args = VocabularySearchRequestArgsSchema
```
ppanero, line 55: This got fixed in master this morning 👼 ... sorry, hope the rebase is not too tricky

invenio_vocabularies/services/querystr.py
```diff
@@ -0,0 +1,55 @@
...
+        """Extract the filtering subtype(s) from query_str.
+
+        Return (<subtypes>, <rest of original query string>).
+        """
+        parts = query_str.split(":", 1)
+        if len(parts) == 1:
+            subtypes = []
+            rest_query_str = query_str
+        else:
+            # Simplification: we can enforce no comma in subtype at
```
ppanero, line 51: I agree on enforcing this, type/subtype should be one word strings... Im even thinking about resource_type and the underscore... so I would say yes to this enforment

invenio_vocabularies/services/querystr.py
```diff
@@ -0,0 +1,55 @@
...
+        """Constructor."""
+        super().__init__(identity=identity, extra_params=extra_params)
+        self.filter_field = filter_field
+
+    def parse(self, query_str):
+        """Parse the query."""
+        subtype_s, query_str = self.extract_subtype_s(query_str)
+        query = super().parse(query_str)
+        if subtype_s:
+            query = query & Q('terms', **{self.filter_field: subtype_s})
```
ppanero, line 38: How do we enforce the mapping being a `keyword`? this would be in rdm-records right?

invenio_vocabularies/version.py
```diff
@@ -13,4 +13,4 @@
 and parsed by ``setup.py``.
 """
 
-__version__ = "0.7.3"
+__version__ = "0.7.4"
```
ppanero, line 16: rebase 😓 v0.7.5 should be now

tests/test_alembic.py
```diff
@@ -36,6 +36,9 @@ def test_alembic(app, database):
     tables = [x.name for x in db.get_tables_for_bind()]
     assert 'vocabularies_metadata' in tables
     assert 'vocabularies_types' in tables
+    assert 'vocabularies_subtypes' in tables
+    # NOTE: The below was created 5 months ago and context unclear, so
```
ppanero, line 40: These are remainings from works on subjects that got stopped. Once affiliations are done, and we start with subjects, this part will be re-done. 

Major: I learnt it not long ago the same way, but we should **not use NOTE** and just leave it as a normal comment if it's not a TODO/FIXME because this keyword (NOTE) has a specific meaning for translations and will be picked up and appear in translation files 😅 I imagine the confusion of translators hehe

I've been removing them when I see them, but I used them quite extensively, let's try to remove them.

[Reference](http://babel.pocoo.org/en/latest/messages.html#translator-comments)

ppanero (approved):
LGTM. If blocking, can merge but:
- Please let's discuss about `gen-migration.sh` I am quite opinionated about not merging that here.
- Remove `NOTE` from comments.

gen-migration.sh
fenekku: Let's remove it for now since it's an extra. I am not sure about the practical side of the gist. I mean `run-tests.sh` is generic, but it's present in each repo. Let's revisit later: https://github.com/inveniosoftware/invenio-vocabularies/issues/60

invenio_vocabularies/alembic/1d3302223885_create_vocabularies_subtype.py
fenekku: They won't be skipped but the next migration will cause a slight headache for the migrator. They could merge them (https://alembic.sqlalchemy.org/en/latest/branches.html#merging-branches), but we might as well make the migrations follow one another and simplify everybody's life. I will "rebase" the migration on top of the latest one (affiliations one).

invenio_vocabularies/records/models.py
fenekku: yes... for now :thinking:

invenio_vocabularies/records/models.py
fenekku: Right now it just means that a different id must be used for a subtype if there is another subtype with desired id but different type :man_shrugging: . In terms of current functionality there is no issue (maybe the current code is not retrieving as fast as it could...), but I am not a DBA. Leaving as is...

invenio_vocabularies/resources/resource.py
fenekku: That was bound to happen with PRs sitting there a while... Rebase is no issue.

invenio_vocabularies/records/models.py
ppanero: I was just wondering if we might end up getting into PK collisions, but if not encountered with MeSH/Subjects which is one of the most complex... I guess we will be good :)

gen-migration.sh
ppanero: Yeay, thanks 👼

invenio_vocabularies/alembic/1d3302223885_create_vocabularies_subtype.py
ppanero: Thanks for the reference, reading it now :)

invenio_vocabularies/services/querystr.py
fenekku: Thanks for highlighting this. I added the missing checks.

tests/test_alembic.py
fenekku: Ha :smile: ! Good thing to know about `NOTE` . I will remove.

fenekku:
Looks like lowest tests are going to time out so merging.

PR #56, opened 2021-06-22
affiliations vocabulary resource layer

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/267
closes inveniosoftware/invenio-rdm-records#328

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/config.py
+++ b/invenio_vocabularies/contrib/affiliations/config.py
@@ -21,28 +21,27 @@ class AffiliationsSearchOptions(SearchOptions):
     """Search options."""
 
     suggest_parser_cls = SuggestQueryParser.factory(
-        fields=[  # TODO: Tweak with name and acronym
-            'id.text^100',
-            'id.text._2gram',
-            'id.text._3gram',
-            'title.en^5',
-            'title.en._2gram',
-            'title.en._3gram',
+        fields=[
+            'name^100',
+            'acronym^20',
+            'title.*^5',
+            'title.*._2gram',
+            'title.*._3gram',
         ],
     )
 
     sort_default = 'bestmatch'
 
-    sort_default_no_query = 'title'
+    sort_default_no_query = 'name'
 
     sort_options = {
         "bestmatch": dict(
             title=_('Best match'),
             fields=['_score'],  # ES defaults to desc on `_score` field
         ),
-        "title": dict(
-            title=_('Title'),
-            fields=['title_sort'],
+        "name": dict(
+            title=_('Name'),
+            fields=['name_sort'],
         ),
         "newest": dict(
             title=_('Newest'),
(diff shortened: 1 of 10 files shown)


invenio_vocabularies/resources/resource.py
```diff
@@ -68,27 +69,6 @@ class VocabulariesResourceConfig(RecordResourceConfig):
     }
 
 
-#
-# Decorators
```
ppanero, line 72: None of these implements anything new with respect to the original ones in flask-resources.

tests/conftest.py
```diff
@@ -25,7 +25,7 @@
 pytest_plugins = ("celery.contrib.pytest", )
 
 
-@pytest.fixture()
+@pytest.fixture(scope="module")
```
ppanero, line 28: default is `function` and we do not need this to be recreated...

tests/contrib/affiliations/test_affiliations_resource.py
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    ]
+    for aff in affiliations:
+        service.create(identity, aff)
+
+    Affiliations.index.refresh()  # Refresh the index
+
+
+def test_affiliations_suggest_sort(
```
ppanero, line 129: I am assuming this is the type of behaviour we want *only* on suggest... wouldn't it be nice on search too?

invenio_vocabularies/contrib/affiliations/config.py
```diff
@@ -21,28 +21,27 @@ class AffiliationsSearchOptions(SearchOptions):
...
-            'id.text^100',
-            'id.text._2gram',
-            'id.text._3gram',
-            'title.en^5',
-            'title.en._2gram',
-            'title.en._3gram',
+        fields=[
+            'name^100',
+            'acronym^20',
+            'title.*^5',
```
ppanero, line 27: In other places we have it as only `en`, i.e. `title.en^5` which will make the suggester only apply to English content... e.g. the [generic vocabulary suggester](https://github.com/inveniosoftware/invenio-vocabularies/blob/31b18084d32c93dbdbc62ff92034fba220a85503/invenio_vocabularies/services/service.py#L41).

If someone wants to test what I mean, you can edit the this by replacing the `*` by `en` and run the test `test_affiliations_suggest_sort` at the bottom. Nothing will be found when `suggest=nucleaire` since it is inside `fr` and not `en`.

I can fix the rest in a separate issue/PR if desired. It would also made a *good first issue*. With this as reference :)

tests/contrib/affiliations/test_affiliations_resource.py
lnielsen: I'm not sure what you mean with the question - perhaps ask me in chat.

PR #58, opened 2021-06-30
services: Cache bug fix in read_all

jrcastro2:


invenio_vocabularies/resources/resource.py
```diff
@@ -52,7 +52,7 @@ class VocabulariesResourceConfig(RecordResourceConfig):
         "type": ma.fields.Str(required=True),
     }
 
-    request_args = VocabularySearchRequestArgsSchema
+    request_search_args = VocabularySearchRequestArgsSchema
```
ppanero, line 55: This needs a bump in records-resources to avoid failing in lowest

ppanero (approved):
LGTM, once the version bump goes in and tests pass

PR #64, opened 2021-07-07
subjects: create subjects specific vocabulary

### LINKED ISSUE #63: Split Subjects into its own Vocabulary
Review list of steps to do like affiliations but for subjects:

**Data layer**

- [x] Define relation in CommonFIeldsMixin.
    - [x] Determine which fields to dump in ES index.
- [x] JSONShema: Do not modify an already released schema, create new version instead.
- [x] JSONSchema: Define or review existing property
    - List or single?
    - Mixed linked with non-linked?
- [x] Mappings adapt to JSONSchema and dumped fields.

**Service layer**

- [x] Schema: Fix metadata schema.
- [x] Facets: Do we need facets? Guillaume: let's skip for now

**Presentation layer**

- [x] Fix REST API serializations
- [x] Fix other serialization formats.

**Tests**

- [x] Add tests that record linking is wokring as expected (e.g. from service layer).
    - Bad and good data.
- [x] Add REST API tests for serialization, facets
- [x] Ensure that I18N is tested properly if required.

Follow structure from:

https://codimd.web.cern.ch/vc6wJipAS66l1XjwKjYdmA?both#

fenekku:
closes #63


invenio_vocabularies/contrib/subjects/jsonschemas/subjects/subject-v1.0.0.json
```diff
@@ -1,21 +1,28 @@
...
   "description": "Subjects from specific vocabularies.",
   "properties": {
-    "identifier": {
-      "description": "Subjects term identifier.",
+    "$schema": {
+      "$ref": "local://definitions-v1.0.0.json#/$schema"
+    },
+    "id": {
+      "description": "Subject unique identifier.",
       "type": "string"
```
ppanero, line 11: there is a definition `identifier`, which in the end is a string. We can use it here.

invenio_vocabularies/contrib/subjects/record_type.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
```
ppanero, line 1: I think we are calling this file `Affiliations.py`, so either we call this `subjects.py` or we rename affiliations. I'm leaning towards `subjects.py`

invenio_vocabularies/contrib/subjects/schema.py
```diff
@@ -1,42 +1,22 @@
...
+from marshmallow import fields
 
+from invenio_vocabularies.services.schema import i18n_strings
 
-class MetadataSchema(Schema):
-    """Basic metadata schema class."""
 
-    class Meta:
-        """Meta class to accept unknown fields."""
+class SubjectSchema(BaseRecordSchema):
```
ppanero, line 17: even if not all the fields are used, we should inherit from `BaseVocabularySchema`. Mainly to keep consistency with the RFC 🤷‍♂️

ppanero:
very tiny comments, a really nice work @fenekku. I'm taking it from here, the only thing needed is an agreement on the actual data model of it. Hopefully with @lnielsen we can set it this morning and merge this before your shift 🚀 

I'm pushing changes in a different commit just to make easier to spot changes, but we should squash before merging hehe

invenio_vocabularies/contrib/subjects/jsonschemas/subjects/subject-v1.0.0.json
```diff
@@ -1,21 +1,28 @@
...
     },
-    "scheme": {
-      "description": "Identifier scheme.",
+    "name": {
```
ppanero, line 13: data mode is tbd, but should we call this `subject`? UI is ready to configure this field name. Name makes sense too

tests/contrib/affiliations/test_affiliations_jsonschema.py
```diff
@@ -82,7 +82,7 @@ def test_fails_acronym(appctx, schema):
 def test_fails_name(appctx, schema):
     data = {
         "$schema": schema,
-        "acronym": 1
+        "name": 1
```
ppanero, line 85: he, nice catch

tests/contrib/subjects/test_subjects_jsonschema.py
```diff
@@ -0,0 +1,85 @@
...
+def fails(data):
+    """Validates affiliation data."""
+    pytest.raises(ValidationError, validates, data)
+    return True
+
+
+def test_valid_full(appctx, schema):
+    data = {
+        "$schema": schema,
+        # Should it always be either "id" or "name"? Or can it be both?
```
ppanero, line 39: can be both... even "should" be both. As far as I understood `id` is the PK in the end, but `name` is the label shown to the user (at least in affiliations).

tests/contrib/subjects/test_subjects_jsonschema.py
```diff
@@ -0,0 +1,85 @@
...
+        "id": "mesh-D000001",
+        "pid": {
+            "pk": 1,
+            "status": "R",
+            "pid_type": "subid",
+            "obj_type": "sub"
+        },
+        "subtype_id": "mesh",
+        # Not a hard requirement, e.g. MeSH doesn't have translations.
```
ppanero, line 48: Do we forsee it in others? @lnielsen otherwise I'd remove and just keep name

tests/contrib/subjects/test_subjects_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+"""Test the subject vocabulary service."""
+
+import pytest
+from invenio_pidstore.errors import PIDAlreadyExists, PIDDeletedError
+from marshmallow.exceptions import ValidationError
+from sqlalchemy.exc import IntegrityError
+
+
+def test_controlled_subject_simple_flow(
```
ppanero, line 18: I think we have a difference on use cases... you are pointing towards "users can add subjects to a subtype" and for affiliations what we did was "free text/custom ones are stored in the record". That means:
- No aggregations possible
- No sharing between records

ppanero:
Subjects have been implemented in the same manner as affiliations. Things to discuss:
- Data model
- Use case of custom subjects being sabed to the db instead of just attach them to a record

invenio_vocabularies/contrib/subjects/record_type.py
ppanero: Saw the pre-existing `subjects.py` I guess it was done like this to avoid removing the old one. I move it to `subjects.py`.

fenekku (commented):
Better late than never?

invenio_vocabularies/contrib/subjects/api.py
```diff
@@ -1,13 +1,14 @@
...
 # details.
 
-"""Vocabulary subjects."""
+"""Subject API."""
 
-from .subjects import subject_record_type
+from .subjects import record_type
 
-Subjects = subject_record_type.record_cls
+Subjects = record_type.record_cls
```
fenekku, line 14: I originally made this one singular because `Record` in api.py is singular too.

invenio_vocabularies/contrib/subjects/jsonschemas/subjects/subject-v1.0.0.json
```diff
@@ -1,22 +1,30 @@
...
-      "description": "Identifier scheme.",
-      "type": "string"
+    "id": {
+      "description": "Subject unique identifier.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
     },
-    "term": {
-      "description": "Subject term value.",
+    "name": {
+      "description": "Alternative to id. Custom free-text keyword.",
```
fenekku, line 16: Like was brought up, I am not sure anymore if we even want to store custom keywords here actually. Just storing them on the record is fine.

fenekku:
@ppanero I squashed everything into 1 commit, so git reset --hard on this branch locally if you need to modify it further (but you shouldn't have to). 

In a separate PR I will remove filtering on generic vocabularies (and related things), but this one is good to be reviewed for separate subject. I will look into loading tomorrow and adapt Northwestern's invenio-subjects-mesh to this new structure.

invenio_vocabularies/cli.py
```diff
@@ -18,7 +18,7 @@
 from invenio_access import any_user
 from invenio_db import db
 
-from invenio_vocabularies.contrib.subjects.subjects import subject_record_type
+from invenio_vocabularies.contrib.subjects.subjects import record_type
```
ppanero, line 21: this import breaks below, it's using the old `subject_record_type`

invenio_vocabularies/contrib/subjects/config.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class SubjectsSearchOptions(SearchOptions):
+    """Search options."""
+
+    suggest_parser_cls = FilteredSuggestQueryParser.factory(
+        filter_field='scheme',
+        fields=[  # suggest fields
+            'subject^100',
+            'subject._2gram',
```
ppanero, line 27: This works for `search_as_you_type` fields. I've added that to v7 subject (not available in v6)

invenio_vocabularies/contrib/subjects/jsonschemas/subjects/subject-v1.0.0.json
```diff
@@ -1,22 +1,28 @@
...
-      "type": "string"
+    "id": {
+      "description": "URI or classification code as identifier - globally unique among all subject schemes.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
     },
-    "term": {
-      "description": "Subject term value.",
-      "type": "string"
+    "pid": {
+      "description": "Still needed?.",
```
ppanero, line 16: yes, for the PIDField to be "re-built" easily if I recall correctly.

invenio_vocabularies/contrib/subjects/resources.py
```diff
@@ -0,0 +1,17 @@
...
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Test the affiliation vocabulary resources."""
+
+from .subjects import record_type
+
+SubjectsResourceConfig = record_type.resource_config_cls
+SubjectsResourceConfig.routes["item"] = "/<path:pid_value>"
```
ppanero, line 15: 🚀

invenio_vocabularies/contrib/subjects/schema.py
ppanero: This still applies. It was agreed we should comply with the RFC when doing affiliations. We can discuss this with Lars on Monday, but I would go for the RFC since is what has been agree for the rest. If we decide to change, we need to change Affiliations too.

invenio_vocabularies/contrib/subjects/schema.py
```diff
@@ -1,42 +1,25 @@
...
-
-        unknown = EXCLUDE
-
-    id = fields.Str()
-    metadata = fields.Nested(MetadataSchema)
-    created = fields.Str()
-    updated = fields.Str()
-    links = Links()
-    revision_id = fields.Integer(dump_only=True)
+    # id in BaseRecordSchema is not required, but I don't see why it shouldn't
```
ppanero, line 20: agree on this.

run-tests.sh
```diff
@@ -19,7 +19,7 @@ function cleanup {
 }
 trap cleanup EXIT
 
-python -m check_manifest --ignore ".*-requirements.txt"
+python -m check_manifest --ignore ".*-requirements.txt,pyproject.toml"
```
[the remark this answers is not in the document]
ppanero: I'm not using pipenv and it's not happening to me 😅 I think we should leave it generic (or add it everywhere, like every module).

tests/contrib/subjects/test_subjects_api.py
```diff
@@ -0,0 +1,74 @@
...
+
+def test_subject_indexing(
+    app, db, es, example_subject, indexer, search_get
+):
+    """Test indexing of a subject."""
+    # Index document in ES
+    assert indexer.index(example_subject)["result"] == "created"
+
+    # Retrieve document from ES
+    print("example_subject.id", example_subject.id)
```
ppanero, line 56: 😅

tests/contrib/subjects/test_subjects_resource.py
```diff
@@ -0,0 +1,137 @@
...
+def test_suggest(client, h, prefix, example_subjects):
+    """Test FilteredSuggestParam."""
+    # No filter
+    res = client.get(f'{prefix}?suggest=abdo', headers=h)
+    assert res.json["hits"]["total"] == 4
+
+    # Single filter
+    res = client.get(f'{prefix}?suggest=MeSH:abdo', headers=h)
+    assert res.status_code == 200
+    assert res.json["hits"]["total"] == 2, res.json
```
ppanero, line 127: why the `, res.json`?

tests/contrib/subjects/test_subjects_resource.py
```diff
@@ -0,0 +1,137 @@
...
+            'subject': 'Abdomen',
+        },
+    ]
+    records = [service.create(identity, s) for s in subjects]
+    Subject.index.refresh()  # Refresh the index
+    return records
+
+
+def test_suggest(client, h, prefix, example_subjects):
+    """Test FilteredSuggestParam."""
```
ppanero, line 119: We should try to test order here, but with this is pretty complex and it is not as easy since is a filtered suggestion...

ppanero (approved):
some minor changes (e.g. one import fix)

ppanero:
I am merging and releasing to unblock things like the frontend. If any points need revisting we can discuss later on/monday and fix in a different PR.

tests/contrib/subjects/test_subjects_resource.py
fenekku: So that on a failed assert, the json is shown. I've found this to be much more efficient in testing/debugging as it directly shows the state of things.

invenio_vocabularies/cli.py
```diff
@@ -18,7 +18,8 @@
 from invenio_access import any_user
 from invenio_db import db
 
-from invenio_vocabularies.contrib.subjects.subjects import subject_record_type
+from invenio_vocabularies.contrib.subjects.subjects import \
+    record_type as subject_record_type
```
fenekku, line 22: Yeah this file needs to be changed. The create subjects vocabulary below is new to me and certainly doesn't work now. I will still do the loading in rdm-records, but I wouldn't mind eventually moving it back here if it makes sense.

Issue #66, opened 2021-07-09
contrib: URLs in subjects id cause problems in resolution


ppanero:
<https://github.com/inveniosoftware/invenio-rdm-records/blob/master/tests/resources/vocabularies/test_subjects_vocabulary.py#L36>

ppanero:
Seems as if the route does not have the `path:` applied and therefore it is not found.
<img width="619" alt="Screenshot 2021-07-12 at 15 40 28" src="https://user-images.githubusercontent.com/6756943/125297475-8d5f6100-e327-11eb-928a-3933fb79cf05.png">

PR #67, opened 2021-07-09
proxies+schemes: add affiliations+subjects proxies and scheme table

### LINKED ISSUE #62: contrib: implement proxies for services/resources
This would affect the way `invenio-rdm-records` access them, e.g. in the affiliations fixture

fenekku:
- closes #62 
- also keeps a generic table (VocabularySubtype) for any vocabulary with subtypes (or schemes) to use to place their metadata in.
  * maybe this only makes sense for subjects, but perhaps grants too? affiliations?
 - where to store services for access can also be discussed as part of this PR
 - also reverts the FilteredSuggester for generic vocabularies


invenio_vocabularies/ext.py
```diff
@@ -33,13 +37,31 @@ def init_app(self, app):
...
         self.service = VocabulariesService(
             config=app.config["VOCABULARIES_SERVICE_CONFIG"],
         )
         self.resource = VocabulariesResource(
             service=self.service,
             config=app.config["VOCABULARIES_RESOURCE_CONFIG"],
         )
+        # Specific vocabularies
+        # - Affiliations
+        self.affiliations_service = AffiliationsService(
```
ppanero, line 50: major: I understand this is needed for the proxies (I am doubting now if we can actually have them...) but:
- Affiliations and subjects should not be registered in invenio-vocabularies. This step should happen in rdm-records. Think that they are "contrib", in this case we want both but there might be potential contribs that we do not want.
- It will break/duplicate with <https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/ext.py>
- If continuing with this, there should be a `init_services` method befor `init_resources`. (see link above)

invenio_vocabularies/records/models.py
```diff
@@ -56,22 +56,22 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
...
-    vocabulary_id = db.Column(
-        db.String,
-        db.ForeignKey(VocabularyType.id, ondelete='CASCADE'),
-        nullable=False,
-    )
-    label = db.Column(db.String)
-    prefix_url = db.Column(db.String)
-    """Any extra metadata is added as columns."""
+    # This is e.g. `subjects`, 'affiliations', ...
+    parent_id = db.Column(db.String, primary_key=True)
```
ppanero, line 71: Should it be an FK? or not because there wont be an entry in the `VocabularyType` table?

invenio_vocabularies/records/models.py
```diff
@@ -56,22 +56,22 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
...
+    name = db.Column(db.String)
```
ppanero, line 72: What's the difference between `id` and `name`?

run-tests.sh
```diff
@@ -20,7 +20,7 @@ function cleanup {
 }
 trap cleanup EXIT
 
-python -m check_manifest --ignore ".*-requirements.txt"
+python -m check_manifest --ignore ".*-requirements.txt,pyproject.toml"
```
ppanero, line 23: I am still not convinced about adding this, it "only" affects those installing local modules via pipenv... From my perspective, either we add it everywhere either has to be added locally somehow.

ppanero (commented):
Assigning @lnielsen to check the resource/service registration. We can discuss on the standup

invenio_vocabularies/records/models.py
fenekku: Not a FK because no entry in VocabularyType table for specific vocabularies.

invenio_vocabularies/records/models.py
fenekku: For MeSH for example:

id: "MeSH"
name: "Medical Subject Headings"
parent_id: "subjects"
uri: "https://www.nlm.nih.gov/mesh/meshhome.html"

run-tests.sh
fenekku: Sorry! This was not meant to be committed. I am only doing it locally. Removing!

invenio_vocabularies/ext.py
lnielsen: Correct. Everything in ``contrib`` is the responsibility of the module using it to register and provides proxies for, hence we cannot create and initialize them here.

invenio_vocabularies/records/models.py
```diff
@@ -56,22 +56,22 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
 
 
 class VocabularySubtype(db.Model):
-    """Vocabulary subtype type model.
+    """Vocabulary subtype model.
```
lnielsen, line 59: This entire model needs to be moved to Subjects contrib and renamed to ``SubjectScheme``. I've opened an issue: https://github.com/inveniosoftware/invenio-vocabularies/issues/69

fenekku: Per discussion in standup: keep but rename to VocabularyScheme. The idea is that we might use it for affiliations at least. Let's see!

fenekku:
@ppanero / @lnielsen This one has been updated with VocabularyScheme and services remark. It can be re-checked. The invenio-rdm-records PR relies on it.

invenio_vocabularies/records/models.py
```diff
@@ -55,30 +55,31 @@ class VocabularyMetadata(db.Model, RecordMetadataBase):
...
+    or separate specific vocabularies.
+
+    It is only used to store metadata for subject schemes for now.
+    But we might store affiliations's schemes or other schemes later.
     """
 
-    __tablename__ = "vocabularies_subtypes"
+    __tablename__ = "vocabularies_schemes"
 
     id = db.Column(db.String, primary_key=True)
```
ppanero, line 70: I wonder if the `SubjectMetadata` table should have a FK to this.

invenio_vocabularies/alembic/4a9a4fd235f8_create_vocabulary_schemes.py
```diff
@@ -0,0 +1,37 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Create vocabulary schemes."""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '4a9a4fd235f8'
+down_revision = '6312f33645c1'
```
ppanero, line 14: comment for reviewers: this point to the affiliations table recipe, so we are all good

invenio_vocabularies/records/models.py
fenekku: Hmm... I interpreted the codimd to say to place the reference to this id in the json of `SubjectsMetadata` and `id` is human readable e.g. `MeSH` so we don't need joins.

invenio_vocabularies/contrib/subjects/services.py
```diff
@@ -9,9 +9,21 @@
...
 from .subjects import record_type
 
 SubjectsServiceConfig = record_type.service_config_cls
 
-SubjectsService = record_type.service_cls
+
+class SubjectsService(record_type.service_cls):
+    """Subjects service."""
+
+    def create_scheme(self, identity, id_, name="", uri=""):
```
fenekku, line 23: More of a convenience for the subjects service than a necessity if we go with supporting schemes more generally (i.e. affiliations would have to implement its own... meh)

invenio_vocabularies/contrib/affiliations/__init__.py
```diff
@@ -1,9 +1,20 @@
...
+
+from .resources import AffiliationsResource, AffiliationsResourceConfig
+from .services import AffiliationsService, AffiliationsServiceConfig
+
+__all__ = [
+    "AffiliationsResource",
+    "AffiliationsResourceConfig",
+    "AffiliationsService",
+    "AffiliationsServiceConfig",
+]
```
fenekku, line 20: I snuck this one in to help on https://github.com/inveniosoftware/invenio-vocabularies/issues/66

Issue #68, opened 2021-07-09
facets not working for subjects


ppanero:
because the labelling class goes to look for the custom ones and those do not exist in the db/es

fenekku:
Facets are not working for resource types either so splitting this off into a separate issue: https://github.com/inveniosoftware/invenio-rdm-records/issues/785

PR #80, opened 2021-08-02
subjects: label facets

fenekku:
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/68
- release: v0.8.1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+Pipfile
+Pipfile.lock
+pyproject.toml
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/invenio_vocabularies/contrib/subjects/facets.py
+++ b/invenio_vocabularies/contrib/subjects/facets.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 Northwestern University.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Subjects services."""
+
+
+class SubjectsLabels:
+    """Fetching of subjects labels for facets."""
+
+    def __call__(self, ids):
+        """Return the mapping when evaluated.
+
+        For subjects, the ids received are actually human-readable and coming
+        from subject and scheme. So we use them as-is!
+        """
+        unique_ids = list(set(ids))
+        return {id_: id_ for id_ in unique_ids}
--- a/invenio_vocabularies/contrib/subjects/facets.py
+++ b/invenio_vocabularies/contrib/subjects/facets.py
@@ -15,8 +15,9 @@ class SubjectsLabels:
     def __call__(self, ids):
         """Return the mapping when evaluated.
 
-        For subjects, the ids received are actually human-readable and coming
-        from subject and scheme. So we use them as-is!
+        In this case, the ids received are actually the vocabulary `scheme`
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -31,6 +31,9 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+Pipfile
+Pipfile.lock
+pyproject.toml
```
ppanero, line 36: If we are gonna add this to every module we should add them in the `cookiecutter-invenio-module` so they are there from the initial commit.

invenio_vocabularies/contrib/subjects/facets.py
```diff
@@ -0,0 +1,22 @@
...
+"""Subjects services."""
+
+
+class SubjectsLabels:
+    """Fetching of subjects labels for facets."""
+
+    def __call__(self, ids):
+        """Return the mapping when evaluated.
+
+        For subjects, the ids received are actually human-readable and coming
```
ppanero, line 18: Is this the case for every subject scheme items? I understand the scheme itself is ok (e.g. MeSH is a human readable) but the items... i.e.

- MeSH: https://raw.githubusercontent.com/galterlibrary/invenio-subjects-mesh/master/invenio_subjects_mesh/vocabularies/subjects_mesh.yaml
- FOS: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/189/files

They both have a URL in the `id` field. I'm wondering then what's the frontend sending as id if it's human readable 🤔

ppanero (commented):
I might be missing some context. So if blocking go ahead. I am wondering about the id *as-is* in the labels.

tests/contrib/subjects/test_subjects_facets.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_subjects_label_duplicates():
+    labels_instance = SubjectsLabels()
+
+    result = labels_instance(["MeSH", "MeSH"])
+
+    assert {"MeSH": "MeSH"} == result
+
+    result = labels_instance(["Abdomen", "Abdomen", "Abdomen"])
```
ppanero, line 21: Wouldn't we receive the URL here? I'm not sure

.gitignore
fenekku: Sure... created here: https://github.com/inveniosoftware/cookiecutter-invenio-module/pull/145

invenio_vocabularies/contrib/subjects/facets.py
fenekku: Ah maybe `ids` is what is tripping up here. These `ids` are actually either the vocabulary `scheme` (top-level) or `subject` (nested level). Not the vocabulary `id`. That's what the comment was about. `ids` as an argument was kept to be *consistent* with the interface of the `Labels` category of classes (e.g. [`VocabularyLabels`](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/services/facets.py#L39)).

What about `label_keys` as a better argument name for both this class and `VocabularyLabels` ?

tests/contrib/subjects/test_subjects_facets.py
fenekku: No, see above.

invenio_vocabularies/contrib/subjects/facets.py
ppanero: Ah I see, thanks. Maybe clarifying it in the docstring is enough (like putting there almost your answer from above).

fenekku: ok I just reworded the existing comment.

fenekku:
Ok merging this one from talk in standup

Issue #84, opened 2021-10-12
contrib: create names (orcid) vocabulary


ppanero:
## Context

Creators and Contributors become a vocabulary, the whole object not just the ORCiD identifier.

## Pre-requisites
- [x] Define a name for the vocabulary: `names` (approved by Lars)
- [x] Define a data model (which fields will the vocabulary contain) https://github.com/inveniosoftware/invenio-vocabularies/issues/83

Note: the following steps of the implementation should follow the same semantics/mechanism used for *subjects* and *affiliations*.

## Data Layer
- [x] Create a package in `invenio_vocabularies/contrib/<name_plural>`
- [x] JSONSchema: ``jsonschemas/<name plural>/<name>-v1.0.0.json``
    - Use defs ``title``, ``description``, ``icon`` from ``invenio_vocabularies/records/jsonschemas/vocabularies/definitions-v1.0.0.json`` (preferably) 
    - Use defs ``id``, ``schema``, ``pid`` from invenio-records-resources.
- [x] Mappings: Add mappings for v6 and v7 following (``mappings/{v6,v7}/<name plural>/<name>-v1.0.0.json``). Names and other attributes might contain non-latin charaters, for searchability we might want to use [asciifolding].(https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html)
- [x] Record type factory: ``<name-plural>.py``:
    - Permissions policy use ``invenio_vocabularies/serivces/permissions.py:PermissionPolicy``
    - Endpoint ``/<name-plural>``
- [x] API/Models: ``api.py`` and ``models.py`` (use from factory)
- [x] Alembic recipe for creating tables (``invenio_vocabularies/alembic``).

## Service layer

- [x] Service schema: ``schema.py``
    - [x] Inherit from ``BaseVocabularySchema``.
    - [x] #95
- [x] Service: ``service.py``
    - [x] Add SearchOptions:
        - [x] Need to check if we inherit from ``VocabularySearchOptions``.
        - [x] suggest_parser_cls may need tweaking
        - [x] Need facets?
        - [x] Need facets labelling? (i.e. will RDM-Records have facets over this vocab)
    - [x] Customize components? Probably need ``DataComponent``, ``PIDComponent``. 
    - [x] Do we use PIDs or UUIDs.

## Resource layer

- [x] Config:
    - [x] May need to define serializer ``"application/vnd.inveniordm.v1+json"`` with associated schema.

[...]

PR #93, opened 2021-10-14
names: add contrib vocabulary

ppanero:
- requires https://github.com/inveniosoftware/invenio-records-resources/issues/289
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/84
- triggers https://github.com/inveniosoftware/invenio-vocabularies/issues/98

**Notes for the reviewers**
- The PID provider posses some complexities, but in order to unlock the loading/cli work and merge the vocabulary, work on the provider/pids will be done as a separate issue in https://github.com/inveniosoftware/invenio-vocabularies/issues/94

### DIFF
--- a/invenio_vocabularies/contrib/names/config.py
+++ b/invenio_vocabularies/contrib/names/config.py
@@ -1,3 +1,4 @@
+
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 CERN.
@@ -6,7 +7,7 @@
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
-"""Vocabulary affiliations configuration."""
+"""Vocabulary names configuration."""
 
 from flask import current_app
 from flask_babelex import lazy_gettext as _
@@ -18,21 +19,20 @@ from werkzeug.local import LocalProxy
 
 from ...services.components import PIDComponent
 
-affiliation_schemes = LocalProxy(
-     lambda: current_app.config["VOCABULARIES_AFFILIATION_SCHEMES"]
+names_schemes = LocalProxy(
+     lambda: current_app.config["VOCABULARIES_NAMES_SCHEMES"]
 )
 
 
-class AffiliationsSearchOptions(SearchOptions):
+class NamesSearchOptions(SearchOptions):
     """Search options."""
 
     suggest_parser_cls = SuggestQueryParser.factory(
         fields=[
             'name^100',
-            'acronym^20',
-            'title.*^5',
-            'title.*._2gram',
-            'title.*._3gram',
+            'family_name^100',
+            'given_name^100',
+            'identifiers.identifier^20',
         ],
     )
 
--- a/invenio_vocabularies/contrib/names/config.py
+++ b/invenio_vocabularies/contrib/names/config.py
@@ -33,6 +33,7 @@ class NamesSearchOptions(SearchOptions):
             'family_name^100',
             'given_name^100',
             'identifiers.identifier^20',
+            'affiliations.name^10',
         ],
     )
 
--- a/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
+++ b/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
@@ -1,0 +1,49 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "local://names/name-v1.0.0.json",
+  "description": "Names vocabulary.",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "$schema": {
+      "$ref": "local://definitions-v1.0.0.json#/$schema"
+    },
+    "id": {
+      "description": "URI or classification code as identifier - globally unique among all names schemes.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
+    "pid": {
+      "$ref": "local://definitions-v1.0.0.json#/internal-pid"
+    },
+    "scheme": {
+      "description": "Identifier of the name scheme.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
+    "name": {"type": "string"},
+    "given_name": {"type": "string"},
+    "family_name": {"type": "string"},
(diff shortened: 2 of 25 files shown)


invenio_vocabularies/contrib/names/schema.py
```diff
@@ -0,0 +1,92 @@
...
+from invenio_records_resources.services.records.schema import BaseRecordSchema
+from marshmallow import Schema, ValidationError, fields, post_load, \
+    validates_schema
+from marshmallow_utils.fields import IdentifierSet, SanitizedUnicode
+from marshmallow_utils.schemas import IdentifierSchema
+
+from .config import names_schemes
+
+
+# FIXME: this is a dup from rdm-records, shall we add it in the aff vocab?
```
slint, line 23: Indeed it's weird that this is not in this module (since it's pretty much a Service schema validation rule for creating affiliations). Maybe, shelve and we see it can be addressed separately.

invenio_vocabularies/contrib/names/schema.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @post_load
+    def calculate_name(self, data, **kwargs):
+        """Update names for person.
+
+        Fill name from given_name and family_name if person.
+        """
+        family_name = data.get("family_name")
+        given_name = data.get("given_name")
+        if family_name and given_name:
```
slint, line 89: Shouldn't this check if there is already `name`, and in that case do nothing?

tests/contrib/names/conftest.py
```diff
@@ -0,0 +1,100 @@
...
+    """Full name data."""
+    return {
+        "name": "Doe, John",
+        "given_name": "John",
+        "family_name": "Doe",
+        "identifiers": [
+            {
+                "identifier": "0000-0001-8135-3489",
+                "scheme": "orcid"
+            }
```
slint, line 78: I would add a GND as well for completeness.

slint (commented):
Minor things regarding search and schema, but otherwise LGTM :+1:

invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "URI or classification code as identifier - globally unique among all names schemes.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
+    "pid": {
+      "$ref": "local://definitions-v1.0.0.json#/internal-pid"
+    },
+    "scheme": {
+      "description": "Identifier of the name scheme.",
+      "$ref": "local://definitions-v1.0.0.json#/identifier"
+    },
```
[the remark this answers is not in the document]
ppanero: The scheme is the "sub-vocabulary" or "vocabulary type". e.g. for subjects we have the global vocabulary, but then there are different inputs.
- Vocabulary: Subjects
    - Scheme: MeSH
    - Scheme: FOS
    - ...
 
Currently for affiliations it would be `RoR` indeed, and for ours `ORCiD`. Unless we do not want to distinguish different sources. It might make deduplication more complicated, but it already is. Knowing where the vocabulary items com from is useful IMO.

invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
```diff
@@ -0,0 +1,49 @@
...
+    "affiliations": {
+      "description": "Affiliations of the person.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "id": {
+            "$ref": "local://definitions-v1.0.0.json#/identifier"
```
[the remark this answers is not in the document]
ppanero: Nope, because it is a vocabulary. So while the affiliation itself will have more fields, the "reference" in the name will be:
- `id` if it is an affiliation existing in the DB
- `name` if a custom one

See for example the affiliation in [`rdm-records`](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json#L2)

TL;DR: They can have multiple identifiers, yes (in their inner data) but they are referenced by one (the PK in the DB, sort of -  pidstore style)

tests/contrib/names/conftest.py
ppanero: added

invenio_vocabularies/contrib/names/schema.py
ppanero: done https://github.com/inveniosoftware/invenio-vocabularies/issues/98

invenio_vocabularies/contrib/names/config.py
```diff
@@ -0,0 +1,67 @@
...
+
+class NamesSearchOptions(SearchOptions):
+    """Search options."""
+
+    suggest_parser_cls = SuggestQueryParser.factory(
+        fields=[
+            'name^100',
+            'family_name^100',
+            'given_name^100',
+            'identifiers.identifier^20',
```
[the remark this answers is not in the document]
ppanero: I agree, I add `affiliations.name`, the `id` it is basically the dumper field of the relation or the name if custom.The `id` is an identifier that might not be searched for (Do we expect people to look by RoR?, the suggesters on ES for affiliations are not meant/optimized for that). 

It is tricky to test and I feel like between `^10` and default (`^1`) there is not much difference. Like the `^100` definitely work, but more granular cases are hard to test.

invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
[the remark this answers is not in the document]
ppanero: To be taken into account when importing actual data in the system. Don't want to subtype/schemed it with `orcid`.

invenio_vocabularies/contrib/names/schema.py
ppanero: IRL:
- do not overwrite `name` 
- validate: either `name` either `family_name` and `given_name`

slint:
Included in #100

PR #96, opened 2021-10-18
names: resolution by identifier 

### LINKED ISSUE #94: contrib: custom pid provider for names
**Problem**
*Name* records are using `RecordIdV2` provider. The format of the generated id is fine, however, the PID type should be different (`nameid`).

**Possible solution*
Edit the PID provider factory to accept a base class, then modify the attribute `pid_type`. This is prefered over creating a custom provider since only said attribute needs to be modified.

**Other questions**
- How do we deal with duplicates if the PID is "random"? This is not a problem only for duplicates coming from different sets (e.g. ORCiD and GND) but also from the same ORCiD dump (testing for the existence of each element before inserting might be expensive).
- How do we implement resolution per id (e.g. resolve an ORCiD). We might want to put the identifiers in a `pids` field, in the same manner, we do for DOIs in the records.

**For the developer**
- Must contain tests at service level

ppanero:
- requires https://github.com/inveniosoftware/invenio-vocabularies/pull/93 therefore only the last commit should be reviewed/taken into account.
- closes #94 

**Changes**
- Modifies PID provider factory to accept a base class
- Adds resolution method at service level: It uses ES in order to make it efficient since we have to search in the body of the name record.
- Adds `/names/<pid_type>/<pid_value` for name record resolution at REST API level

**Open question**
- How do we deal with duplicates in the loading process? Do we want to store one PID per identifier (We might be exploding PIDStore and might want to split in tables...)

### DIFF
--- a/invenio_vocabularies/contrib/names/names.py
+++ b/invenio_vocabularies/contrib/names/names.py
@@ -35,7 +35,9 @@ record_type = RecordTypeFactory(
     # Data layer
     pid_field_kwargs={
         "create": False,
-        "provider": RecordIdProviderV2,
+        "provider": PIDProviderFactory.create(
+            pid_type='nameid', base_cls=RecordIdProviderV2
+        ),
         "context_cls": BaseVocabularyPIDFieldContext,
     },
     schema_version="1.0.0",
--- a/invenio_vocabularies/contrib/names/names.py
+++ b/invenio_vocabularies/contrib/names/names.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary names."""
+
+from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
+from invenio_records.systemfields import RelationsField
+from invenio_records_resources.factories.factory import RecordTypeFactory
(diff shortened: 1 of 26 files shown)


invenio_vocabularies/contrib/names/services.py
```diff
@@ -8,8 +8,49 @@
...
 from .names import record_type
 
 NamesServiceConfig = record_type.service_config_cls
 
-NamesService = record_type.service_cls
+
+class NamesService(record_type.service_cls):
+    """Name service."""
+
+    def resolve(self, identity, id_, id_type):
```
slint, line 23: Shouldn't this be using the DB instead of ES, to resolve to a single Name record? Each "formal" identifier we have in the metadata could be a PID that gets resolved.

invenio_vocabularies/contrib/names/names.py
```diff
@@ -35,7 +35,9 @@
     # Data layer
     pid_field_kwargs={
         "create": False,
-        "provider": RecordIdProviderV2,
+        "provider": PIDProviderFactory.create(
+            pid_type='nameid', base_cls=RecordIdProviderV2
```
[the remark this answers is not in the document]
slint: It's actually 6, my bad

invenio_vocabularies/contrib/names/services.py
ppanero: IRL: A decission needs to be taken on making a `PIDSField` (systemfield). This would allow to do DB resolution, instead of going through ES.

slint:
Included in #100

PR #123, opened 2022-01-07
contrib: add orcid http reader for names

ppanero:
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/126

Example of programmatic usage (which is the flow of datastream aprox):

```
>>> from invenio_vocabularies.contrib.names.datastreams import OrcidHTTPReader, OrcidXMLTransformer
>>> reader = OrcidHTTPReader(orcid_id="0000-0001-6759-6273", test_mode=False)
>>> OrcidXMLTransformer().apply(reader.read())
{'given_name': 'Pablo', 'family_name': 'Panero', 'identifiers': [{'scheme': 'orcid', 'identifier': 'https://orcid.org/0000-0001-6759-6273'}], 'affiliations': []}
```

### DIFF
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -8,11 +8,26 @@
 
 """Names datastreams, transformers, writers and readers."""
 
+import requests
 from invenio_access.permissions import system_identity
 from invenio_records.dictutils import dict_lookup
 
 from ...datastreams.errors import TransformerError
+from ...datastreams.readers import SimpleHTTPReader
 from ...datastreams.transformers import XMLTransformer
+
+
+class OrcidHTTPReader(SimpleHTTPReader):
+    """ORCiD HTTP Reader."""
+
+    def __init__(self, *args, test_mode=True, **kwargs):
+        """Constructor."""
+        if test_mode:
+            origin = "https://sandbox.orcid.org/{id}"
+        else:
+            origin = "https://orcid.org/{id}"
+
+        super().__init__(origin, *args, **kwargs)
 
 
 class OrcidXMLTransformer(XMLTransformer):
@@ -42,7 +57,7 @@ class OrcidXMLTransformer(XMLTransformer):
         }
 
         try:
-            employments =  dict_lookup(
+            employments = dict_lookup(
                 record, "activities-summary.employments.affiliation-group"
             )
             if isinstance(employments, dict):
@@ -61,6 +76,11 @@ class OrcidXMLTransformer(XMLTransformer):
             pass
 
         return entry
+
+
+VOCABULARIES_DATASTREAM_READERS = {
+    "orcid-http": OrcidHTTPReader,
+}
 
 
 VOCABULARIES_DATASTREAM_TRANSFORMERS = {
--- a/invenio_vocabularies/datastreams/readers.py
+++ b/invenio_vocabularies/datastreams/readers.py
@@ -11,6 +11,7 @@
 import re
 import tarfile
 
+import requests
 import yaml
 
 
@@ -54,3 +55,29 @@ class TarReader(BaseReader):
                 if member.isfile() and match:
                     content = archive.extractfile(member).read()
                     yield content
+
+
+class SimpleHTTPReader(BaseReader):
+    """Simple HTTP Reader."""
+
+    def __init__(
+        self, origin, id=None, ids=None, content_type=None, *args, **kwargs
+    ):
+        """Constructor."""
+        assert id or ids
+        self._ids = ids if ids else [id]
+        self.content_type = content_type
+        super().__init__(origin, *args, **kwargs)
+
+    def read(self):
+        """Reads a yaml file and returns a dictionary per element."""
+        headers = {"Accept": self.content_type}
+
+        for id_ in self._ids:
+            url = self._origin.format(id=id_)
+            resp = requests.get(url, headers=headers)
+            if resp.status_code != 200:
+                # todo add logging/fail
+                pass
+
+            yield resp.content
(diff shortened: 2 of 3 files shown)


tests/contrib/names/test_names_datastreams.py
```diff
@@ -58,11 +42,48 @@ def xml_entry():
...
-    assert expected_from_xml == transformer.apply(xml_entry)
+    assert expected_from_xml == transformer.apply(bytes_xml_entry)
+
+
+class MockResponse():
+    text = XML_ENTRY_DATA
+    status_code = 200
+
+
+@patch('requests.get', side_effect=lambda url, headers: MockResponse())
```
ppanero, line 86: not sure if this is the best way to test it... but is the sanest (i.e. not making external requests to orcid each time we test).

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,13 +8,43 @@
...
+    def __init__(self, orcid_id, test_mode=True, *args, **kwargs):
+        """Constructor."""
+        if test_mode:
+            origin = f"https://sandbox.orcid.org/{orcid_id}"
+        else:
+            origin = f"https://orcid.org/{orcid_id}"
+
+        super().__init__(origin, *args, **kwargs)
+
+    def read(self):
```
ppanero, line 32: was doubting if `orcid_id` should be a parameter here instead of the constructor. But then we would create a reading loop with diff parameters each time, which breaks the API of the other readers.

i.e. I don't like the fact of creating an object per ORCiD to fetch, but it is what's meant for.

tests/resources/test_tasks_resources.py
```diff
@@ -0,0 +1,77 @@
...
+        },
+        "transformers": [{"type": "test"}],
+        "writers": [{"type": "test"}]
+    }
+
+    breakpoint()
+    resp = client.post(
+        "/vocabularies/tasks", headers=h, data=json.dumps(task_config)
+    )
+    assert resp.status_code == 202
```
ppanero, line 74: for some reason I am getting:
```
{'status': 400, 'message': 'A validation error occurred.', 'errors': [{'field': 'type', 'messages': ['Missing data for required field.']}]}
```

invenio_vocabularies/services/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+"""Celery tasks."""
+
+from celery import shared_task
+from flask import current_app
+
+from ..datastreams.factories import DataStreamFactory
+
+
+@shared_task(ignore_result=True)
+def launch_async_vocab_task(config):
```
slint, line 17:
```suggestion
def process_datastream(config):
```

invenio_vocabularies/services/service.py
```diff
@@ -196,3 +205,21 @@ def read_many(self, identity, type, ids, fields=None, **kwargs):
...
+        the PIDs, having a sub-service and a manager. If persistance is
+        added UoW should be used.
+        """
+        self.require_permission(identity, "manage")
+        task_config, _ = self.task_schema.load(
+            data,
+            context={"identity": identity},  # FIXME: is this needed
+            raise_errors=True
+        )
+        launch_async_vocab_task(task_config)
```
slint, line 222:
```suggestion
        launch_async_vocab_task.delay(task_config)
```

To call it async. But maybe there are troubles with passing the `Identity` object in there... I guess it goes back to the discussion we had about how to "serialize" an identity.

invenio_vocabularies/contrib/names/datastreams.py
ppanero: base signature accepts args and kwargs, could have gone there...

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,13 +8,43 @@
...
 
 from ...datastreams.errors import TransformerError
+from ...datastreams.readers import BaseReader
 from ...datastreams.transformers import XMLTransformer
 
 
+class OrcidHTTPReader(BaseReader):
+    """ORCiD HTTP Reader."""
+
+    def __init__(self, orcid_id, test_mode=True, *args, **kwargs):
```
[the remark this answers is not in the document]
ppanero: I "agree", but is an orcid reader... why wouldn't we hardcode the endpoint? if we use `origin` we might as well expose `headers` and make it a `SimpleHTTPReader`... no?

ppanero:
@slint I moved the Tasks REST API to https://github.com/inveniosoftware/invenio-vocabularies/pull/124 so this is more self contained (was getting messy to keep the tree clean when implementing two features in one PR). I'm moving the corresponding comments there. This is ready for re-review.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -54,3 +55,29 @@ def read(self):
                 if member.isfile() and match:
                     content = archive.extractfile(member).read()
                     yield content
+
+
+class SimpleHTTPReader(BaseReader):
```
ppanero, line 60: IMO this is now more flexible, allows for:
- different content-types
- different endpoints, as long as the url (origin) is a string to be formatted. I did it this way because concatenating might not always work (I bet some endpoints might not have the id at the end...)
- `id` and `ids` was not to force an array, I think it is a "prettier" API this way, accepting one or the other (favouring `ids`).

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,13 +8,28 @@
...
 
 from ...datastreams.errors import TransformerError
+from ...datastreams.readers import SimpleHTTPReader
 from ...datastreams.transformers import XMLTransformer
 
 
+class OrcidHTTPReader(SimpleHTTPReader):
+    """ORCiD HTTP Reader."""
+
+    def __init__(self, *args, test_mode=True, **kwargs):
```
ppanero, line 23: I still kept this to enforce the URL, otherwise we just use the `SimpleHTTPReader`.

PR #124, opened 2022-01-11
Tasks REST API

ppanero:
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/125

**I am not convinced of the implementation**, it looks a bit ad-hoc (no uow, adding `task_schema` to the config, and so on). IMO, it should be a separate resource (like PIDs in `invenio-rdm-records`) but this serves as PoC.

@slint might be missing a test at service level... but is indirectly tested by the resource since the tasks are run eagerly in tests.

### DIFF
--- a/invenio_vocabularies/services/permissions.py
+++ b/invenio_vocabularies/services/permissions.py
@@ -21,6 +21,3 @@ class PermissionPolicy(RecordPermissionPolicy):
     can_update = [SystemProcess()]
     can_delete = [SystemProcess()]
     can_manage = [SystemProcess()]
-
-    # Type permissions
-    can_manage = [SystemProcess()]
--- a/invenio_vocabularies/services/schema.py
+++ b/invenio_vocabularies/services/schema.py
@@ -29,10 +29,33 @@ class BaseVocabularySchema(BaseRecordSchema):
 
 
 class VocabularySchema(BaseVocabularySchema):
-    """Service schema for vocabulary records.."""
+    """Service schema for vocabulary records."""
 
     props = fields.Dict(
         allow_none=False, keys=fields.Str(), values=fields.Str()
     )
     tags = fields.List(SanitizedUnicode())
     type = fields.Str(attribute='type.id', required=True)
+
+
+class DatastreamObject(Schema):
+    """Datastream object (reader, transformer, writer)."""
+
+    type = fields.Str(required=True)
+    args = fields.Dict(keys=fields.Str(), values=fields.Field)
+
+
+class TaskSchema(Schema):
+    """Service schema for vocabulary tasks."""
+
+    reader = fields.Nested(DatastreamObject, required=True)
+    transformers = fields.List(
+        fields.Nested(DatastreamObject),
+        validate=validate.Length(min=1),
+        required=False,
+    )
+    writers = fields.List(
+        fields.Nested(DatastreamObject),
+        validate=validate.Length(min=1),
+        required=True,
+    )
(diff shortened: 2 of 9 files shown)


invenio_vocabularies/services/schema.py
```diff
@@ -29,10 +29,33 @@ class BaseVocabularySchema(BaseRecordSchema):
...
         allow_none=False, keys=fields.Str(), values=fields.Str()
     )
     tags = fields.List(SanitizedUnicode())
     type = fields.Str(attribute='type.id', required=True)
+
+
+class DatastreamObject(Schema):
+    """Datastream object (reader, transformer, writer)."""
+
+    type = fields.Str(required=True)
```
ppanero, line 44: Not sure if we should validate this here (kind of difficult though, since it's based on `reader/transformer/writer`) or if it's enough actually the underlying API's job to raise an appropriate exception

invenio_vocabularies/services/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+"""Celery tasks."""
+
+from celery import shared_task
+from flask import current_app
+
+from ..datastreams.factories import DataStreamFactory
+
+
+@shared_task(ignore_result=True)
+def launch_async_vocab_task(config):
```
ppanero, line 17:
```suggestion
def process_datastream(config):
```

invenio_vocabularies/services/service.py
```diff
@@ -196,3 +205,21 @@ def read_many(self, identity, type, ids, fields=None, **kwargs):
...
+        the PIDs, having a sub-service and a manager. If persistance is
+        added UoW should be used.
+        """
+        self.require_permission(identity, "manage")
+        task_config, _ = self.task_schema.load(
+            data,
+            context={"identity": identity},  # FIXME: is this needed
+            raise_errors=True
+        )
+        launch_async_vocab_task(task_config)
```
ppanero, line 222:
```suggestion
        launch_async_vocab_task.delay(task_config)
```

To call it async. But maybe there are troubles with passing the `Identity` object in there... I guess it goes back to the discussion we had about how to "serialize" an identity.

ppanero (commented):
review from @slint, initially done in https://github.com/inveniosoftware/invenio-vocabularies/pull/123

invenio_vocabularies/services/service.py
ppanero: It is okay, the identity comes from a JSON file or from a request payload and therefore is a string. It would be either imported either use the `system_process` as default. This was one of the big questions in the earlier PRs of `datastreams`, decided to default to `system_identity` (see https://github.com/inveniosoftware/invenio-vocabularies/blob/names/invenio_vocabularies/datastreams/writers.py#L51 and discussions on the corresponding PR)

invenio_vocabularies/services/schema.py
ppanero: I think is okay to validate up to this point (i.e. that there is a `type` and `args`, and no other field). Then, the _args_ are dependent on the `reader/transformer/writer` so that is up to them. However, with no `type` we cannot even choose one, so it is mandatory.

invenio_vocabularies/services/permissions.py
```diff
@@ -21,6 +21,3 @@ class PermissionPolicy(RecordPermissionPolicy):
     can_update = [SystemProcess()]
     can_delete = [SystemProcess()]
     can_manage = [SystemProcess()]
-
-    # Type permissions
-    can_manage = [SystemProcess()]
```
ppanero, line 26: duplicated from above

slint (approved):

PR #129, opened 2022-01-13
cli: implement update command and adapt service writer

### LINKED ISSUE #86: cli: vocabulary update
Implement a command to update the content of a vocabulary, i.e.

`invenio vocabularies update resourcetype`

Must take into account the vocabulary nature and support:
- updating from a local file
- updating from a remote file
- prioritized update
- filtered update (e.g. update based on a condition, like having the CERN ROR)
- update only a subtype (e.g. subjects-mesh)

Most of these functionalities should be available through loader mixins.

Potential `ignore` and `force` arguments to support updating entries in the `BaseFixture._load_vocabulary` function.

ppanero:
closes #86 

Implements changes on the `ServiceWrite` to allow updating a record if it exists, by doing this the rest of the pipeline (readers, transformers) stay the same.

### DIFF
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -10,12 +10,15 @@
 
 import requests
 from invenio_access.permissions import system_identity
+from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_records.dictutils import dict_lookup
+from marshmallow import ValidationError
 
 from ...datastreams import StreamEntry
-from ...datastreams.errors import TransformerError
+from ...datastreams.errors import TransformerError, WriterError
 from ...datastreams.readers import SimpleHTTPReader
 from ...datastreams.transformers import XMLTransformer
+from ...datastreams.writers import ServiceWriter
 
 
 class OrcidHTTPReader(SimpleHTTPReader):
@@ -76,7 +79,51 @@ class OrcidXMLTransformer(XMLTransformer):
         except Exception:
             pass
 
-        return StreamEntry(entry)
+        stream_entry.entry = entry
+        return stream_entry
+
+
+class NamesServiceWriter(ServiceWriter):
+    """Names service writer."""
+
+    def __init__(self, *args, scheme_id="orcid", **kwargs):
+        """Constructor."""
+        self._scheme_id = scheme_id
+        super().__init__(*args, **kwargs)
+
+    def _entry_id(self, entry):
+        """Get the id from an entry."""
+        for identifier in entry.get("identifiers"):
+            if identifier.get("scheme") == self._scheme_id:
+                return identifier["identifier"]
+
+    def _resolve(self, id_):
+        """Resolve an entry given an id."""
+        return self._service.resolve(
+            self._identity, id_=id_, id_type=self._scheme_id
+        )
+
+    def write(self, stream_entry, *args, **kwargs):
+        """Writes the input entry using a given service."""
+        entry = stream_entry.entry
+        try:
+            vocab_id = self._entry_id(entry)
+            # it is resolved before creation to avoid duplicates since
+            # the pid is recidv2 not e.g. the orcid
+            current = self._resolve(vocab_id)
+            if not self._update:
+                raise WriterError(
+                    [f"Vocabulary entry already exists: {entry}"]
+                )
+            updated = dict(current.to_dict(), **entry)
+            return StreamEntry(
+                self._service.update(self._identity, current.id, updated)
+            )
+        except PIDDoesNotExistError:
+            return StreamEntry(self._service.create(self._identity, entry))
+
+        except ValidationError as err:
+            raise WriterError([{"ValidationError": err.messages}])
 
 
 VOCABULARIES_DATASTREAM_READERS = {
@@ -89,6 +136,13 @@ VOCABULARIES_DATASTREAM_TRANSFORMERS = {
 }
 """ORCiD Data Streams transformers."""
 
+
+VOCABULARIES_DATASTREAM_WRITERS = {
+    "names-service": NamesServiceWriter,
+}
+"""ORCiD Data Streams transformers."""
+
+
 DATASTREAM_CONFIG = {
     "reader": {
         "type": "tar",
@@ -100,7 +154,7 @@ DATASTREAM_CONFIG = {
         {"type": "orcid-xml"}
     ],
     "writers": [{
-        "type": "service",
+        "type": "names-service",
         "args": {
             "service_or_name": "rdm-names",
             "identity": system_identity,
--- a/invenio_vocabularies/datastreams/writers.py
+++ b/invenio_vocabularies/datastreams/writers.py
@@ -12,9 +12,11 @@ from pathlib import Path
 
 import yaml
 from invenio_access.permissions import system_identity
+from invenio_pidstore.errors import PIDAlreadyExists
 from invenio_records_resources.proxies import current_service_registry
 from marshmallow import ValidationError
 
+from .datastreams import StreamEntry
 from .errors import WriterError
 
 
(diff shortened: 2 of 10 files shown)


invenio_vocabularies/datastreams/writers.py
```diff
@@ -53,12 +54,19 @@ def __init__(self, service_or_name, identity, *args, **kwargs):
...
+    def write(self, stream_entry, force=False, *args, **kwargs):
+        """Writes the input entry using a given service.
+
+        :param force: if True update if exists, otherwise fail.
+        """
         try:
             result = self._service.create(self._identity, stream_entry.entry)
         except ValidationError as err:
             raise WriterError([{"ValidationError": err.messages}])
+        except PIDAlreadyExists as err:
```
ppanero, line 66: Would it be more desirable to do a `read()` instead of letting it raise a PK violation? I know in python we should "except" by default, but in this case we are generating an error at DB level.

The `read` is required anyway if `force == True` since we will require it to add changes to the record body.

tests/contrib/names/test_names_datastreams.py
```diff
@@ -89,3 +124,23 @@ def test_orcid_http_reader(_, bytes_xml_entry):
...
     assert bytes_xml_entry.entry == results[0].entry
     assert not results[0].errors
+
+
+def test_names_service_writer_create(app, expected_from_xml):
+    writer = NamesServiceWriter("rdm-names", system_identity)
+    record = writer.write(StreamEntry(expected_from_xml))
+    record = record.to_dict()
+
+    assert dict(record, **expected_from_xml) == record
```
ppanero, line 134: comment: something weird is happening here, the `IdentifierSet` is stripping out the `scheme` and the url:

```shell
E         Differing items:
E         {'identifiers': [{'identifier': 'https://orcid.org/0000-0001-8135-3489', 'scheme': 'orcid'}]} != {'identifiers': [{'identifier': '0000-0001-8135-3489', 'scheme': 'orcid'}]}
```

tests/contrib/names/test_names_datastreams.py
```diff
@@ -89,3 +125,57 @@ def test_orcid_http_reader(_, bytes_xml_entry):
...
+
+
+def test_names_service_writer_duplicate(app, expected_from_xml):
+    writer = NamesServiceWriter("rdm-names", system_identity)
+    _ = writer.write(stream_entry=StreamEntry(expected_from_xml))
+    with pytest.raises(WriterError) as err:
```
ppanero, line 141: Unless we _resolve_ every time (which is not expensive since it goes to ES), there is not DB deduplication because we use `recidv2` and we do not store PIDs for identifiers (metadata only).

invenio_vocabularies/datastreams/writers.py
```diff
@@ -53,15 +54,38 @@ def __init__(self, service_or_name, identity, *args, **kwargs):
...
+
+    def write(self, stream_entry, force=False, *args, **kwargs):
+        """Writes the input entry using a given service.
+
+        :param force: if True update if exists, otherwise fail.
+        """
+        entry = stream_entry.entry
         try:
-            result = self._service.create(self._identity, stream_entry.entry)
+            result = self._create(entry) if not force else self._update(entry)
```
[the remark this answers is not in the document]
slint: Thinking about it again, would it make sense to replace the `force` parameter with something like `update=True`? Then the logic goes as follows in `write(...)`:

```python
def write(self, stream_entry, *args, **kwargs):
    try:
        result = self._service.create(self._identity, stream_entry)
    except PIDAlreadyExists as ex:
        if self.update:
            # See https://github.com/inveniosoftware/invenio-pidstore/blob/8f6e99787da5a07d38712f7eb16146608304cdb9/invenio_pidstore/errors.py#L23-L24
            vocab_id = (ex.pid_type, ex.pid_value)
            current = self._service.read(self._identity, vocab_id)
            updated = dict(current.to_dict(), **stream_entry)
            result = self._service.update(self._identity, vocab_id, stream_entry)
        else:
            # Pass?
    return result
```

invenio_vocabularies/datastreams/writers.py
```diff
@@ -53,15 +54,38 @@ def __init__(self, service_or_name, identity, *args, **kwargs):
...
+        """Updates an entry and creates if it does not exist."""
+        vocab_id = (entry["type"], entry["id"])
+        try:
+            current = self._service.read(self._identity, vocab_id)
+            updated = dict(current.to_dict(), **entry)
+            return self._service.update(self._identity, vocab_id, updated)
+        except PIDDoesNotExistError:
+            return self._service.create(self._identity, entry)
+
+    def write(self, stream_entry, force=False, *args, **kwargs):
```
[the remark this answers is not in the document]
ppanero: I agree with this. I opted by the `force` flag at a `write` level because it seemed more natural to me to give the choice per `record` instead of a `writer` arg (i.e. update all or create all). But it is true that given the use cases we have for `datastreams`/`writers` a writer arg works and pollute less the APIs

tests/contrib/names/test_names_datastreams.py
ppanero: Fixed, it was due to the `IdentifierScheme` saving a normalized (non-URL) value. Then the subsequent searches and comparissons would be comparing the URL to the normalized value.

tests/contrib/names/test_names_datastreams.py
ppanero: Fixed by implementing a customized `write` method.

invenio_vocabularies/datastreams/writers.py
```diff
@@ -53,15 +54,38 @@ def __init__(self, service_or_name, identity, *args, **kwargs):
...
+        except PIDAlreadyExists:
+            raise WriterError(
+                [f"Vocabulary entry already exists: {entry}"]
+            )
+
+    def _update(self, entry):
+        """Updates an entry and creates if it does not exist."""
+        vocab_id = (entry["type"], entry["id"])
+        try:
+            current = self._service.read(self._identity, vocab_id)
```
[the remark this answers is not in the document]
ppanero: Done

invenio_vocabularies/datastreams/writers.py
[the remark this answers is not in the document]
ppanero: Done (left as `write`, since there would be no difference among the two functions.

tests/test_cli.py
```diff
@@ -129,3 +135,19 @@ def test_process(app, names_tar_file, names_service):
...
     assert list(results.hits)[0]["identifiers"][0]["identifier"] == orcid
+
+
+@patch.object(NamesService, "read")
+@patch.object(NamesService, "update")
+@patch.object(NamesService, "create")
+def test_update_cmd(p_create, p_update, p_read, app, names_tar_file):
+    with app.app_context():
+        runner = CliRunner()
+        result = runner.invoke(
```
ppanero, line 146: fails always with:
```
'Usage: vocabularies [OPTIONS] COMMAND [ARGS]...\n\nError: Could not locate a Flask application. You did not provide the "FLASK_APP" environment variable, and a "wsgi.py" or "app.py" module was not found in the current directory.\n'
```

Any ideas? I've checked `invenio-base` tests but do not help much...

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -76,7 +79,49 @@ def apply(self, stream_entry, **kwargs):
...
+            if identifier.get("scheme") == self._scheme_id:
+                return identifier["identifier"]
+
+    def _resolve(self, id_):
+        """Resolve an entry given an id."""
+        return self._service.resolve(
+            self._identity, id_=id_, id_type=self._scheme_id
+        )
+
+    def write(self, stream_entry, *args, **kwargs):
```
slint, line 106: Maybe add a comment (or issue to shelve) in order to remember that we have "reversed" the order here because of the PID handling issues (i.e. try to update first and then create).

slint (approved):
Just "documentation" of a specific design choice, but otherwise it's good to merge :)

PR #140, opened 2022-01-21
[BLOCKED] datastreams: implement asynchronous writer

### LINKED ISSUE #139: writers: support async
Writing is a blocking operation. An `AsyncWriter` could be implemented.

It would receive a "subwriter" as a parameter. Then it would lunch write operations with said "subwriter" asynchronously (celery tasks)

ppanero:
closes #139

### DIFF
--- a/invenio_vocabularies/datastreams/tasks.py
+++ b/invenio_vocabularies/datastreams/tasks.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Data Streams Celery tasks."""
+
+from celery import shared_task
+
+from ..datastreams import StreamEntry
+from ..datastreams.factories import WriterFactory
+
+
+@shared_task(ignore_result=True)
+def write_entry(writer, entry):
+    """Write an entry.
+
+    :param writer: writer configuration as accepted by the WriterFactory.
+    :param entry: dictionary, StreamEntry is not serializable.
+    """
+    writer = WriterFactory.create(config=writer)
+    writer.write(StreamEntry(entry))
--- a/invenio_vocabularies/datastreams/writers.py
+++ b/invenio_vocabularies/datastreams/writers.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2021-2022 CERN.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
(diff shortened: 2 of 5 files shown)


invenio_vocabularies/datastreams/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+from ..datastreams import StreamEntry
+from ..datastreams.factories import WriterFactory
+
+
+@shared_task(ignore_result=True)
+def write_entry(writer, entry):
+    """Write an entry.
+
+    :param writer: writer configuration as accepted by the WriterFactory.
+    :param entry: dictionary, StreamEntry is not serializable.
```
ppanero, line 22: Note this. `StreamEntry` is not serializable so a bit of juggling is needed.

invenio_vocabularies/datastreams/writers.py
```diff
@@ -110,3 +111,21 @@ def write(self, stream_entry, *args, **kwargs):
...
+
+class AsyncWriter(BaseWriter):
+    """Writes the entries asynchronously (celery task)."""
+
+    def __init__(self, writer, *args, **kwargs):
+        """Constructor.
+
+        :param writer: writer to use.
+        """
+        self._writer = writer
```
ppanero, line 124: Accepts a writer configuration, that will be sent to the `WriterFactory` later on. I'm not convinced about the name of this parameter. Alternatives, which I dont fully like either are:
- `writer_conf`, `writer_config`, `w_conf`...
- `config`, `conf`...

Everything is too generic....

ppanero:
@slint it is ready for review. note that names is not writen async right now, but if we agree on this implementation is just a matter of changing this [config](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/names/datastreams.py#L139).

invenio_vocabularies/datastreams/writers.py
```diff
@@ -110,3 +111,21 @@ def write(self, stream_entry, *args, **kwargs):
             yaml.safe_dump([stream_entry.entry], file)
 
         return stream_entry
+
+
+class AsyncWriter(BaseWriter):
```
slint, line 116: From what I understand, this approach is to use this writer as a container for other writer configs.

I was thinking if it would be possible to have a `delay` kwarg in `BaseWriter`, which if `True` would achieve the same result... But from what I see that's pretty difficult to handle, given `__init__` overrides, etc. 

My fear is that we end up with a very complex datastream config structure, where especially in YAML it's easy to make indentation mistakes, which as a result would eventually lead to more difficult UX/DX.

Other approaches to discuss:

- Move this up to the DataStream, i.e. in `BaseDataStream.write(...)`, do something like `if writer.delay: ...call task...`
- Preserve the original writer config/init-params when initializing writers?

ppanero: I agree with you. However, changes are not trivial.

In both cases we would need to implement a way to either save the configuration (_type_ + _args_) or make writers JSON serializable. Otherwise, we won't be able to send them to the task. IMO this fits more at data stream level, i.e. _the writers writes_, but is the data stream who decides _how/when_.

slint:
Rebased and included in https://github.com/inveniosoftware/invenio-vocabularies/pull/357

PR #152, opened 2022-03-07
datastreams: make base classes abstract

### LINKED ISSUE #133: naming: `Base` objects with functionality
In the case of datastreams there is a _base_ class that has an implementation that in many cases suffice. In addition, it is not overwritable.
See examples of cases in `factories.py`. I'm wondering if there should be similar to e.g. the `readers`

- `BaseDataStream` with the skeleton and constructor
- `DataStream` with the current implementation

Note that this also happens with `BaseFixture`

ppanero:
closes #133

### DIFF
--- a/tests/resources/test_tasks_resources.py
+++ b/tests/resources/test_tasks_resources.py
@@ -19,49 +19,50 @@ from invenio_vocabularies.datastreams.transformers import BaseTransformer
 from invenio_vocabularies.datastreams.writers import BaseWriter
 
 
+class TestReader(BaseReader):
+    """Test reader."""
+
+    # no need for self since it is patched
+    def read(*args, **kwargs):
+        """Yields empty dict."""
+        yield StreamEntry(1)
+
+
+class TestTransformer(BaseTransformer):
+    """Test transformer."""
+
+    # no need for self since it is patched
+    def apply(stream_entry, *args, **kwargs):
+        """Forwards entry."""
+        return stream_entry
+
+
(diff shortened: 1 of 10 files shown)


tests/resources/test_tasks_resources.py
```diff
@@ -74,10 +75,21 @@ def test_task_creation(
...
     }
 
-    resp = client.post(
-        "/vocabularies/tasks", headers=h, data=json.dumps(task_config)
-    )
-    assert resp.status_code == 202
-    p_read.assert_called()
-    p_apply.assert_called()
-    p_write.assert_called()
+    with patch.object(
```
ppanero, line 78: not a fan of chaining ctx managers... but I didn't find any other way... open to suggestions 😇

slint: I think you can include multiple comma-separated ctx managers as well in one `with ...` like:

```python
with patch.object(...) as p_read, \
        patch.object(...) as p_apply, \
        patch.object(...) as p_write:
    resp = ...
```

In [Python 3.10 you can even have parentheses](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers) to avoid the ugly backslashes :)

slint (approved):

tests/resources/test_tasks_resources.py
ppanero: Thanks! done, without parenthesis to give support to older than 3.9

Issue #148, opened 2022-03-04
datastreams: funders dump


ppanero:
Process (read items) from ~~CrossRef FunderRef dump, available [here](https://www.crossref.org/services/funder-registry/) --> [Gitlab link to file](https://gitlab.com/crossref/open_funder_registry) (The CSV file might be easier to process, if we need simplified data, e.g. country - not always available)~~ the ROR dump (published [on Zenodo](https://zenodo.org/communities/ror-data)).

Some sub-tasks:

- [x] Piping readers implementation
- [x] `ZipReader`, to be able to read the ROR dump ZIP directly
- [x] `JSONReader` to allow splitting a JSON array into entries
- [x] `FunderTransformer` from the ROR dump to Funder service schema
  - Schema documented [at ROR docs](https://ror.readme.io/docs/ror-data-structure).
- Some logic can be reused from [Invenio OpenAIRE loaders](https://github.com/inveniosoftware/invenio-openaire/blob/master/invenio_openaire/loaders.py).
- Also, check out the old [Zenodo's CLI command](https://github.com/zenodo/zenodo/blob/b27daab5825f310739bc46ca49c9454f65b72ed0/zenodo/modules/fixtures/cli.py#L101). In InvenioRDM it should be done via datastreams (and for compatibility with the *old fixtures*).

PR #153, opened 2022-03-08
datastreams: funders readers and transformers

ctrl-alt-soneca:
requires https://github.com/inveniosoftware/invenio-vocabularies/pull/155
closes https://github.com/inveniosoftware/invenio-vocabularies/issues/148

Example imported from the dump:

```json
{
        "pid": "04j757h98",
        "name": "Victoria University",
        "created": "2022-03-21T16:11:03.049346+00:00",
        "id": "a73cad48-a30b-4cc7-b4aa-366a201a2029",
        "country": "AU",
        "title": {},
        "links": {
          "self": "https://127.0.0.1:5000/api/funders/04j757h98"
        },
        "updated": "2022-03-21T16:11:03.059741+00:00",
        "identifiers": [
          {
            "scheme": "isni",
            "identifier": "0000 0001 0396 9544"
          },
          {
            "scheme": "doi",
            "identifier": "10.13039/501100001784"
          },
          {
            "scheme": "grid",
            "identifier": "grid.1019.9"
          }
        ],
        "revision_id": 2
      }
```

Rebasing multiple PRs has created a bit of conflicts, should be solved once the required PRs are merged. Tests are passing in local:

Screenshot 2022-03-21 at 17 26 20

### DIFF
--- a/invenio_vocabularies/cli.py
+++ b/invenio_vocabularies/cli.py
@@ -18,6 +18,7 @@ from invenio_access.permissions import system_identity
 from invenio_pidstore.errors import PIDDeletedError, PIDDoesNotExistError
 from invenio_records_resources.proxies import current_service_registry
 
+from .contrib.funders.datastreams import DATASTREAM_CONFIG as funders_ds_config
 from .contrib.names.datastreams import DATASTREAM_CONFIG as names_ds_config
 from .datastreams import DataStreamFactory
 
@@ -30,7 +31,17 @@ def get_config_for_ds(vocabulary, filepath=None, origin=None):
             with open(filepath) as f:
                 config = yaml.safe_load(f).get(vocabulary)
         if origin:
-            config["reader"]["args"]["origin"] = origin
+            config["readers"][0]["args"]["origin"] = origin
+
+        return config
+
+    if vocabulary == "funders":
+        config = deepcopy(funders_ds_config)
+        if filepath:
+            with open(filepath) as f:
+                config = yaml.safe_load(f).get(vocabulary)
+        if origin:
+            config["readers"][0]["args"]["origin"] = origin
 
         return config
 
@@ -49,7 +60,7 @@ def vocabularies():
 def _process_vocab(config, num_samples=None):
     """Import a vocabulary."""
     ds = DataStreamFactory.create(
-        reader_config=config["reader"],
+        readers_config=config["readers"],
         transformers_config=config.get("transformers"),
         writers_config=config["writers"],
     )
--- a/invenio_vocabularies/config.py
+++ b/invenio_vocabularies/config.py
@@ -12,7 +12,8 @@
 import idutils
 from flask_babelex import lazy_gettext as _
 
-from .datastreams.readers import TarReader, YamlReader
+from .datastreams.readers import CSVReader, JsonReader, TarReader, \
+    YamlReader, ZipReader
 from .datastreams.transformers import XMLTransformer
 from .datastreams.writers import ServiceWriter, YamlWriter
 from .resources.resource import VocabulariesResourceConfig
@@ -51,8 +52,16 @@ VOCABULARIES_AFFILIATION_SCHEMES = {
 
 VOCABULARIES_FUNDER_SCHEMES = {
     **VOCABULARIES_IDENTIFIER_SCHEMES,
+    "doi": {
+        "label": _("DOI"),
+        "validator": idutils.is_doi
+    },
 }
 """Funders allowed identifier schemes."""
+
+VOCABULARIES_FUNDER_DOI_PREFIX = "10.13039"
+"""DOI prefix for the identifier formed with the FundRef id."""
+
 
 VOCABULARIES_NAMES_SCHEMES = {
     "orcid": {
@@ -74,8 +83,11 @@ VOCABULARIES_NAMES_SCHEMES = {
 """Names allowed identifier schemes."""
 
 VOCABULARIES_DATASTREAM_READERS = {
+    "csv": CSVReader,
+    "json": JsonReader,
     "tar": TarReader,
     "yaml": YamlReader,
+    "zip": ZipReader,
 }
 """Data Streams readers."""
 
--- a/invenio_vocabularies/contrib/funders/datastreams.py
+++ b/invenio_vocabularies/contrib/funders/datastreams.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Funders datastreams, transformers, writers and readers."""
+
+from flask_babelex import lazy_gettext as _
+from idutils import normalize_ror
+from invenio_access.permissions import system_identity
+
+from ...datastreams.errors import TransformerError
+from ...datastreams.transformers import BaseTransformer
+from ...datastreams.writers import ServiceWriter
+from .config import funder_fundref_doi_prefix, funder_schemes
+
+
+class FundersServiceWriter(ServiceWriter):
+    """Funders service writer."""
+
+    def _entry_id(self, entry):
+        """Get the id from an entry."""
+        return entry["pid"]
+
+
+class RORTransformer(BaseTransformer):
+    """Transforms a JSON ROR record into a funders record."""
+
+    def apply(self, stream_entry, **kwargs):
+        """Applies the transformation to the stream entry."""
+        record = stream_entry.entry
+        funder = {}
+
+        funder["pid"] = normalize_ror(record.get("id"))
+        if not funder["pid"]:
+            raise TransformerError(
+                _(f"Id not found in ROR entry."))
+
+        funder["name"] = record.get("name")
+        if not funder["name"]:
+            raise TransformerError(_(f"Name not found in ROR entry."))
+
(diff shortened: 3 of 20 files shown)


invenio_vocabularies/cli.py
```diff
@@ -30,7 +31,17 @@ def get_config_for_ds(vocabulary, filepath=None, origin=None):
             with open(filepath) as f:
                 config = yaml.safe_load(f).get(vocabulary)
         if origin:
-            config["reader"]["args"]["origin"] = origin
+            config["readers"][0]["args"]["origin"] = origin
+
+        return config
+
+    if vocabulary == "funders":
```
ppanero, line 38: action: create an issue for the creation of a "factory" and remove the above `FIXME`

invenio_vocabularies/config.py
```diff
@@ -77,5 +78,6 @@
 VOCABULARIES_DATASTREAM_WRITERS = {
     "service": ServiceWriter,
     "yaml": YamlWriter,
+    "csv": PrintWriter,
```
ppanero, line 81:
```suggestion
    "print": PrintWriter,
```

invenio_vocabularies/contrib/funders/datastreams.py
```diff
@@ -0,0 +1,28 @@
...
+        {
+            "type": "csv",
+            "args": {
+                "csv_options": None,
+                "as_dict": True,
+                "mode": 'r',
+            }
+        }
+    ],
+    # FIXME: add transformers and writers
```
ppanero, line 23: action: fix config (can be done once transformers are implemented)

invenio_vocabularies/datastreams/datastreams.py
```diff
@@ -24,14 +24,14 @@ def __init__(self, entry, errors=None):
 class DataStream:
     """Data stream."""
 
-    def __init__(self, reader, writers, transformers=None, *args, **kwargs):
+    def __init__(self, readers, writers, transformers=None, *args, **kwargs):
         """Constructor.
 
         :param reader: the reader object.
```
ppanero, line 30:
```suggestion
        :param readers: an ordered list of readers.
```

invenio_vocabularies/datastreams/datastreams.py
```diff
@@ -57,6 +57,39 @@ def process(self, *args, **kwargs):
             else:
                 yield self.write(transformed_entry)
 
+    def _original_compose(self, *funcs):
```
ppanero, line 60: minor: remove if unused

invenio_vocabularies/datastreams/factories.py
```diff
@@ -68,10 +68,14 @@ class DataStreamFactory:
 
     @classmethod
     def create(
-        cls, reader_config, writers_config, transformers_config=None, **kwargs
+        cls, readers_config, writers_config, transformers_config=None, **kwargs
     ):
         """Creates a data stream based on the config."""
-        reader = ReaderFactory.create(reader_config)
+        # reader = ReaderFactory.create(reader_config)
```
ppanero, line 74:
```suggestion
```

invenio_vocabularies/datastreams/readers.py
```diff
@@ -87,4 +90,58 @@ def read(self):
...
+                if not member.is_dir() and match:
+                    yield archive.open(member)
+
+
+class JsonReader(BaseReader):
+    """JSON object reader."""
+
+    def read(self, item=None, **kwargs):
+        """Reads (loads) a json object and yields its items."""
+        entries = json.load(item)
```
ppanero, line 121: To be consistent with other readers
```suggestion
        filepath = item or self._origin
        entries = json.load(filepath)
```

invenio_vocabularies/datastreams/writers.py
```diff
@@ -108,3 +108,16 @@ def write(self, stream_entry, *args, **kwargs):
             yaml.safe_dump([stream_entry.entry], file)
 
         return stream_entry
+
+
+class PrintWriter(BaseWriter):
```
ppanero, line 113: major: was this added for testing resons? If yes, I think we should remove it (also from the config above). Otherwise, it should print on the console (currently saving to a yaml file 🤔). This brings me to:

If we are going to print, wouldnt it be better to implement a logging write that is flexible enough? (there was an issue about it but I cant find it).

My take: remove `PrintWriter` seems only used for tests/ad-hoc.

invenio_vocabularies/cli.py
ppanero: https://github.com/inveniosoftware/invenio-vocabularies/issues/164

invenio_vocabularies/datastreams/datastreams.py
ppanero: removed

invenio_vocabularies/datastreams/readers.py
ppanero: Fixed

invenio_vocabularies/datastreams/datastreams.py
mitsosf: Fixed

invenio_vocabularies/datastreams/factories.py
mitsosf: Fixed

invenio_vocabularies/datastreams/writers.py
ppanero: removed

invenio_vocabularies/datastreams/readers.py
```diff
@@ -87,4 +96,61 @@ def read(self):
...
+                if not member.is_dir() and match:
+                    yield archive.open(member)
+
+
+class JsonReader(BaseReader):
+    """JSON object reader."""
+
+    def read(self, item=None, **kwargs):
+        """Reads (loads) a json object and yields its items."""
+        file = item if item else open(self._origin, mode='r')
```
slint, line 127: Is there a way maybe to use a `with ...` clause here (just to avoid the "manual" `file.close()` at the end). Also maybe we make a convention so that it's the responsibility of the previous reader to handle the open/close?

invenio_vocabularies/datastreams/datastreams.py
```diff
@@ -57,6 +57,30 @@ def process(self, *args, **kwargs):
...
+            current_gen_func = _gen_funcs.pop(0)
+            for item in current_gen_func(piped_item):
+                try:
+                    # exhaust iterations of subsequent generators
+                    if _gen_funcs:
+                        yield from pipe_gen(_gen_funcs, piped_item=item)
+                    # there is no subsequent generator, return the current item
+                    else:
+                        yield StreamEntry(item)
+                except ReaderError as err:
```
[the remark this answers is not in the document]
ppanero: it is "yielding" an entry which contains an error. So **all** should be processed, then it is up to the `datastream.process()` to handle that. i.e. [here](https://github.com/inveniosoftware/invenio-vocabularies/pull/153/files/1881e93b892692b2268e3e9663373014f1f834db..7f26518d7126346dd113eb447d681908d7bdce83#diff-4757dd09a819e8a8ed000c77053d4e6f7c6bf20f7448e82729904c2b3419277dR50) where I see we are not doing it 😅. Will fix and add a test for it ;)

invenio_vocabularies/datastreams/readers.py
```diff
@@ -87,4 +96,61 @@ def read(self):
...
+        entries = json.load(file)
+        for entry in entries:
```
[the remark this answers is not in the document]
ppanero: we thought we were assuming always array, but if it's not the case then yes we need to check type.

invenio_vocabularies/datastreams/readers.py
ppanero: there is but it would duplicate the inner code, will try to refactor.

ppanero: The base class implements now with the `with` ctx manager the opening of the file if it's not being piped through item. Some readers like Zip and Tar have to overwrite this.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -8,41 +8,54 @@
...
         self._origin = origin
+        self._mode = mode
 
     @abstractmethod
-    def read(self, *args, **kwargs):
-        """Reads the content from the origin.
-
-        Yields `StreamEntry` objects.
-        """
+    def _read(self, fp, *args, **kwargs):
```
ppanero, line 38: or `iterate`? i.e. `read` opens/reads but this function is the one actually looping... 🤔

invenio_vocabularies/datastreams/readers.py
```diff
@@ -76,15 +95,87 @@ def __init__(
         self.content_type = content_type
         super().__init__(origin, *args, **kwargs)
 
-    def read(self):
-        """Reads a yaml file and returns a dictionary per element."""
+    def _read(self, fp, *args, **kwargs):
```
slint, line 98: The `fp` variable name loses a bit of its meaning here. In that case, maybe dropping it an using something like `url` is better (since we're overriding `read` as well),

invenio_vocabularies/datastreams/readers.py
slint: Agree, any of the below would be possible:

-  `_iter`
- `_iterate`
- `iter`

I somewhat dislike that this starts looking very much like a Python iterator class... Though our use-case does differ because we do the piping etc.

ppanero: done, choose `_iter`. Because it is an iterator + its internal, should not be called from outside `read`.

PR #155, opened 2022-03-08
Funders vocabulary

alejandromumo:
- requires https://github.com/inveniosoftware/invenio-records-resources/pull/308
- requires https://github.com/inveniosoftware/invenio-records-resources/pull/311
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/147


invenio_vocabularies/contrib/funders/mappings/v6/funders/funder-v1.0.0.json
```diff
@@ -0,0 +1,55 @@
...
+          "type": "keyword",
+          "index": "false"
+        },
+        "created": {
+          "type": "date"
+        },
+        "updated": {
+          "type": "date"
+        },
+        "uuid": {
```
ppanero, line 16: (this is more of a exploratory comment)

question: what does this field get as value? and why is it needed? If I recall correctly the `ObjectId` in Elasticsearch is has the same value than the record UUID (which gets linked to the PID value via PIDStore). When denormalizing a vocabulary record into a record we add a `@v` value in order to be able to version changes and update correctly. Is this UUID value needed in order to not query again when performing the denormalization?

(We can debug this together, I am just noting this here to not forget)

invenio_vocabularies/contrib/funders/models.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
```
ppanero, line 3:
```suggestion
# Copyright (C) 2022 CERN.
```

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -9,6 +9,7 @@
 """Funders schema."""
 
 from flask_babelex import lazy_gettext as _
+from invenio_records_resources.services.records.schema import BaseRecordSchema
 from marshmallow import Schema, fields, validate
 from marshmallow_utils.fields import SanitizedUnicode
```
ppanero, line 15: copyright to 2021-2022

tests/contrib/funders/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+        ]
+    }
+
+
+@pytest.fixture(scope="function")
+def funder_full_data():
+    """Full funder data."""
+    return {
+        "scheme": "doi",
+        "id": "fund",
```
ppanero, line 46: moderate: this should potentially fail. The scheme is set to `doi` but the id value is not a DOI. We are not imposing any type of validation on this, which makes me think if we should:
a) Use `IdentifierScheme`, i.e. multiple inheritance (this could cause problems with `id` and `identifier`
b) Use only at validation time

For now this question can be shelved, because it will be completely different if decide to go for `identifiers` (plural). We have to wait for the datamodel to be ready.

tests/contrib/funders/test_funders_jsonschema.py
```diff
@@ -0,0 +1,81 @@
...
+            "pid_type": "funderid",
+            "obj_type": "funder"
+        },
+        "scheme": "Test scheme"
+    }
+
+    assert validates(data)
+
+
+def test_valid_empty(appctx, schema):
```
ppanero, line 53: moderate: shouldn't there be? for the "relations" there should not, but at service level we should not allow insertion of empty funder records.

tests/contrib/funders/test_funders_schema.py
```diff
@@ -29,10 +55,10 @@ def test_valid_minimal():
     assert valid_minimal == FunderRelationSchema().load(valid_minimal)
 
 
-def test_invalid_no_name():
-    invalid_no_name = {
-        "identifier": "10.5281/zenodo.9999999",
+def test_invalid_minimal():
```
ppanero, line 58:
```suggestion
def test_invalid_no_id():
```
WDYT?

tests/contrib/funders/test_funders_schema.py
```diff
@@ -10,16 +10,42 @@
 import pytest
 from marshmallow import ValidationError
 
-from invenio_vocabularies.contrib.funders.schema import FunderRelationSchema
+from invenio_vocabularies.contrib.funders.schema import FunderRelationSchema, \
+    FunderSchema
```
ppanero, line 16: moderate: it ties up with one of the comments above, but there should be a test to check the actual `id`/`scheme` value format.

tests/contrib/funders/test_funders_service.py
```diff
@@ -0,0 +1,82 @@
...
+    # Fail to retrieve it
+    # - db
+    pytest.raises(PIDDeletedError, service.read, identity, id_)
+    # - search
+    res = service.search(
+        identity, q=f"id:{id_}", size=25, page=1)
+    assert res.total == 0
+
+
+# QUESTION is it needed?
```
ppanero, line 67: Yes, we basically have to test that we do not allow duplicates to get in the system.

tests/contrib/funders/test_funders_service.py
```diff
@@ -0,0 +1,82 @@
...
+def test_pid_already_registered(
+    app, db, service, identity, funder_full_data
+):
+    """Recreating a record with same id should fail."""
+    service.create(identity, funder_full_data)
+    pytest.raises(
+        PIDAlreadyExists, service.create, identity, funder_full_data)
+
+
+# QUESTION it should fail because we set it to strict, right?
```
ppanero, line 77: It should fail way before reaching Elasticsearch, since the [marshmallow schema is "strict" by default](https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields).

ppanero (commented):
Very nice work! I tried to be verbose since it is the first PR (so, sorry for the herd of comments in advance). We can discuss all these in a call, it might be easier :) But also feel free to continue on the resource layer since it is kind of independent.

💪 🚀

invenio_vocabularies/contrib/funders/api.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
```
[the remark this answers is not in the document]
alejandromumo: fixed

invenio_vocabularies/contrib/funders/mappings/v7/funders/funder-v1.0.0.json
```diff
@@ -0,0 +1,77 @@
...
+            "type": "keyword"
+          }
+        }
+      },
+      "title": {
+        "type": "object",
+        "dynamic": true,
+        "properties": {
+          "en": {
+            "type": "text"
```
ppanero, line 71: moderate: missing `search_as_you_type` (only for v7). See [affiliations](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/affiliations/mappings/v7/affiliations/affiliation-v1.0.0.json#L76) as example.

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -8,9 +8,16 @@
...
 from marshmallow import Schema, fields, validate
-from marshmallow_utils.fields import SanitizedUnicode
+from marshmallow_utils.fields import IdentifierSet, SanitizedUnicode
+from marshmallow_utils.schemas import IdentifierSchema
+
+from ...services.schema import BaseVocabularySchema
+from .config import funder_schemes
 
 
 class FunderRelationSchema(Schema):
```
ppanero, line 23: Major: This field is missing an `id` field. Otherwise, the "relation" won't be saved. The way the relation schema work is:

For example, the user wants to add 2 funders to it's record. One exists in our system with id `funder-id-1`, the other does not exist and its name is `JDoe funding agency`. The record would be sent (and stored in DB) as:

```json
{ ...
    "funders": [
        {"id": "funder-id-1"},
        {"name": "JDoe funding agency"}
    ]
}
```

Then, internally the record would be denormalized. Therefore the result (for the end user) would be:

```json
{ ...
    "funders": [
        {"id": "funder-id-1", "name": ..., "title":..., "scheme"...},
        {"name": "JDoe funding agency"}
    ]
}
```

tests/contrib/funders/conftest.py
ppanero: This is fixed now that there is an inner `identifiers` field. You can resolve the comment.

tests/contrib/funders/test_funders_api.py
```diff
@@ -0,0 +1,94 @@
...
+# details.
+
+"""Funders API tests."""
+
+from functools import partial
+
+import pytest
+from invenio_indexer.api import RecordIndexer
+from invenio_search import current_search_client
+from jsonschema import ValidationError as SchemaValidationError
```
ppanero, line 16: question: I know I did this rename in `names` (According to git blame 😅) but I do not see the reason why... In some cases it might clash with `Marshmallow.ValidationError` but is not the case here... WDYT?

tests/contrib/funders/test_funders_resource.py
```diff
@@ -0,0 +1,179 @@
...
+    res = client_with_credentials.get(f"{prefix}/{id_}", headers=h)
+    assert res.status_code == 410
+
+
+def test_funders_update(
+    client_with_credentials, example_funder, funder_full_data, h, prefix
+):
+    """Test a successful update."""
+    id_ = example_funder.id
+    new_name_ = "updated"
```
ppanero, line 162:
```suggestion
    new_name = "updated"
```

We put `_` after `id` because it is a reserved python keyword, for other variables is not needed. Ref https://docs.python.org/3/library/functions.html#id

invenio_vocabularies/contrib/funders/config.py
```diff
@@ -0,0 +1,63 @@
...
+        ),
+        "newest": dict(
+            title=_('Newest'),
+            fields=['-created'],
+        ),
+        "oldest": dict(
+            title=_('Oldest'),
+            fields=['created'],
+        ),
+    }
```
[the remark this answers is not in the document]
alejandromumo: Seems good to me 👍

invenio_vocabularies/contrib/affiliations/schema.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2021 CERN.
+# Copyright (C) 2021-2022 CERN.
```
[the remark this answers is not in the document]
alejandromumo: it was totally my mistake here 😅

tests/contrib/funders/test_funders_resource.py
alejandromumo: thanks for the tip!

tests/contrib/funders/test_funders_api.py
alejandromumo: I agree, while testing the API it won't probably clash with any `Marshmallow` exception since `Marshmallow` it is not used at this level (e.g. we're not (de)serialising data).

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -20,5 +27,21 @@ class FunderRelationSchema(Schema):
         required=True,
```
ppanero, line 27: minor: sorry didn't catch this earlier, shouldn't be required here (so we can set it to an "fk" funder)

tests/contrib/funders/test_funders_schema.py
```diff
@@ -10,29 +10,85 @@
...
+def test_invalid_empty_name():
+    invalid_empty_name = {
+        "id": "10.5281/zenodo.9999999",
+        "name": ""
+    }
+    with pytest.raises(ValidationError):
+        data = FunderSchema().load(invalid_empty_name)
+
+
+def test_invalid_id_format():
```
ppanero, line 45: very very minor: Im not sure if all these tests are needed, or are responsibilities of other modules... We can sit down and think about this on Monday.

ppanero (commented):
LGTM! Not approving because is missing the `pid` part, blocked by https://github.com/inveniosoftware/invenio-vocabularies/issues/156

invenio_vocabularies/contrib/funders/config.py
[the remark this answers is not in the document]
ppanero: will be tackeld in https://github.com/inveniosoftware/invenio-vocabularies/issues/159

invenio_vocabularies/contrib/funders/config.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+service_components = [
+    # Order of components are important!
+    DataComponent,
+    PIDComponent,
+]
```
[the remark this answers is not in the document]
ppanero: will be tackled in https://github.com/inveniosoftware/invenio-vocabularies/issues/160

invenio_vocabularies/contrib/funders/mappings/v6/funders/funder-v1.0.0.json
ppanero: https://github.com/inveniosoftware/invenio-vocabularies/issues/161

tests/contrib/funders/test_funders_schema.py
ppanero: I have made some changes to this file, very minor:
- It is good to test fields that are defined here (e.g. `country`). However, others like `identifiers` that come from marshmallow-utils (our module, not 3rd party) we tend to trust and avoid testing many times.
- The relation schema was missing validation.

tests/contrib/awards/test_awards_schema.py
```diff
@@ -68,8 +68,6 @@ def test_invalid_no_number():
 
 FUNDER = {
     "name": "Someone",
-    "identifier": "10.5281/zenodo.9999999",
```
ppanero, line 71: This will fix tests until the grants vocabulary is implemented.

ppanero (commented):
Missing `ModelPIDField` usage, but the vocabulary per se is implemented! 🚀

tests/contrib/funders/test_funders_service.py
```diff
@@ -0,0 +1,87 @@
...
+        identity, q=f"pid:{id_}", size=25, page=1)
+    assert res.total == 0
+
+
+def test_pid_already_registered(
+    app, db, service, identity, funder_full_data
+):
+    """Recreating a record with same id should fail."""
+    service.create(identity, funder_full_data)
+    pytest.raises(
```
ppanero, line 79: @slint, there are seveal things that are puzzling me.
1- Even though the `pid` field has an `unique` constraint, here it is not failing
2- Resolution is not working (`NoResultFound`).

I actually checked the DB after the `service.create` calls, the table is empty 😮 😵

invenio_vocabularies/contrib/funders/config.py
```diff
@@ -0,0 +1,65 @@
...
+        "oldest": dict(
+            title=_('Oldest'),
+            fields=['created'],
+        ),
+    }
+
+
+service_components = [
+    # Order of components are important!
+    DataComponent,
```
ppanero, line 64: `PIDComponent` is not needed since it goes "in-table"

invenio_vocabularies/alembic/3d362de4926a_create_funders_table.py
```diff
@@ -0,0 +1,56 @@
+#
+# This file is part of Invenio.
```
ppanero, line 2: Need re-making since now we have two more columns.

tests/contrib/funders/test_funders_service.py
```diff
@@ -0,0 +1,92 @@
...
+    # not-ideal cleanup
+    item._record.delete(force=True)
+
+
+def test_pid_already_registered(
+    app, service, identity, funder_full_data, example_funder
+):
+    """Recreating a record with same id should fail."""
+    # example_funder does the first creation
+    # FIXME: It is not a "PIDAlreadyExists" should we catch and re-raise
```
ppanero, line 82: @slint what do we do about this?

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -8,17 +8,95 @@
...
+    country = SanitizedUnicode()
+    identifiers = IdentifierSet(fields.Nested(
+        partial(
+            IdentifierSchema,
+            allowed_schemes=funder_schemes,
+            identifier_required=False
+        )
+    ))
+
+    # TODO: the 3 hooks below could be extracted to a PIDField
```
ppanero, line 65: @slint Should we create said field? I do not recall what did you said on the call this morning sorry

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -8,17 +8,95 @@
...
+        """Loads the pid.
+
+        If it does not exist, it tries to extract it from the identifiers.
+        """
+        if data.get("pid"):
+            return data
+
+        identifiers = data.get("identifiers", [])
+        for identifier in identifiers:
+            # FIXME: make scheme configurable
```
ppanero, line 77: Needs fixing, maybe with a config var (similar to `SCHEMES` in affiliations)

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -8,17 +8,96 @@
...
+            if identifier["scheme"] == "ror":
+                data["pid"] = identifier["identifier"]
+                break
+
+        return data
+
+    @post_load
+    def kepp_pid_on_create(self, data, **kwargs):
```
slint, line 85:
```suggestion
    def keep_pid_on_create(self, data, **kwargs):
```

invenio_vocabularies/contrib/funders/schema.py
slint: I'm a bit hesitant actually about allowing "flexibility" in this since we might end up with "duplicate" entries (under different identifiers, etc.).

Maybe just keeping the first `if data.get("pid"):` is the safest, and removes complexity, i.e. you have to explicitly pass the `pid` value when you create the record and that's the one and only way.

invenio_vocabularies/contrib/funders/schema.py
```diff
@@ -8,17 +8,96 @@
...
+        """Remove the PID if already registered.
+
+        If the pid is not removed it would be injected into the DB JSON column.
+        Since the field is only popped on_create by PIDComponent.
```
slint, line 89: Not a big fan of relying on the schema to keep the DB JSON "clean". Maybe there's something we can configure/change in the PIDComponent that would help?

invenio_vocabularies/contrib/funders/schema.py
ppanero: I 100% agree, I added it there because we somehow talked about it. Will remove.

invenio_vocabularies/contrib/funders/funders.py
```diff
@@ -0,0 +1,53 @@
...
+
+record_type = RecordTypeFactory(
+    "Funder",
+    # Data layer
+    pid_field_cls=ModelPIDField,
+    pid_field_kwargs={
+        "model_field_name": "pid",
+    },
+    model_cls_attrs={
+        # cannot set to nullable=False because it would fail at
```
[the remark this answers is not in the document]
ppanero: Sadly, yes...

invenio_vocabularies/contrib/funders/schema.py
ppanero: Yes, we could potentially create a custom PIDComponent and pop it in the update hook but:
- It is still at service level (marshamllow and components both are service level)
- when we talked about this we agreed to do marshmallow 😅 but yeah, I dont like it either, just didnt find another solution.

invenio_vocabularies/contrib/funders/schema.py
ppanero: As a result, the whole `load_pid` hook gets removed. The first `if` clause performs no changes on the incoming data.

invenio_vocabularies/alembic/676dd587542d_create_funders_vocabulary_table.py
```diff
@@ -0,0 +1,67 @@
...
+            nullable=False
+        ),
+        sa.Column(
+            'updated',
+            sa.DateTime().with_variant(mysql.DATETIME(fsp=6), 'mysql'),
+            nullable=False
+        ),
+        sa.Column(
+            'id',
+            sqlalchemy_utils.types.uuid.UUIDType(),
```
ppanero, line 38: This should have a `length=16`, same as teh `ChoiceType` below a `length=1`. However, it is not compatible with alembic :/

invenio_vocabularies/contrib/funders/schema.py
ppanero: resolving this comment because it boils down to the one below: i.e. where do we managed the pid/id? in the component or in marshmallow

invenio_vocabularies/contrib/funders/schema.py
slint: Would it make sense to make `pid` a `dump_only` field maybe and validate separately (as done now kind of)?

invenio_vocabularies/alembic/676dd587542d_create_funders_vocabulary_table.py
slint: I see that in the `names` recipe it's the same. Isn't it fine to leave it like this?

invenio_vocabularies/contrib/funders/schema.py
ppanero: not really, because you still need it during loading (i.e. for creation). We would need something like `dump_only_on_updates`, which is what we achieve here.

invenio_vocabularies/alembic/676dd587542d_create_funders_vocabulary_table.py
ppanero: yes, it is. I just added a comment in case of doubt.

invenio_vocabularies/services/components.py
```diff
@@ -49,3 +49,12 @@ def create(self, identity, data=None, record=None, **kwargs):
...
         self.service.record_cls.pid.create(record)
+
+
+class ModelPIDComponent(PIDComponent):
+    """Model PID registration component."""
+
+    def update(self, identity, data=None, record=None, **kwargs):
+        """Update an existing record."""
+        record.pop("pid", None)
+        record.pop("id", None)
```
slint, line 60: Question: where does the `id` field come from in the record metadata?

slint (approved):

invenio_vocabularies/services/components.py
ppanero: Because it is an update and the `result_item.data` does a `marshmallow.dump`, which has to contain the id.

PR #157, opened 2022-03-10
[PoC] names: extract pid to a systemfield

### LINKED ISSUE #156: contrib: remove PIDs usage for big vocabularies

ppanero:
closes #156 

There are two approaches to this issue:

### DB column

- Data would be stored in the DB separately. Could be used to resolve a names record.

### SystemField

(The implemented one in this PR)

- Data would not be duplicated, it would be extracted from a configurable field (`identifiers`).
- Resolution would be implemented using the ES resolver (currently at service level, need to move to the Provider, **need to check how to do to keep separation of concerns**).

### Problems

**Both approaches** make now the "identifiers" field mandatory. Either to set it to a DB column or to extract it from metadata with a systemfield, it needs to be there. (I do understand that names use `recidv2` but funders would use the ror so we have this problem).

cc @slint


tests/contrib/names/test_names_api.py
```diff
@@ -101,5 +100,5 @@ def test_name_pid(app, db, example_name):
     """Test name pid creation."""
     name = example_name
 
-    assert name.pid.pid_value
-    assert name.pid.pid_type == "names"
+    assert name.pid == "0000-0001-8135-3489"
```
ppanero, line 103: Many tests might fail. This is an example to showcase it actually works as expected (creates PID value from string `identifiers`, resolves)

ppanero:
@slint resolvers and systemfields can easily move to invenio-records-resources, Im not 100% sure about the provider in this case, since it is quite specific to contrib vocabularies and the `identifiers` field.

invenio_vocabularies/contrib/names/api.py
```diff
@@ -1,13 +1,18 @@
...
 
+from ...records.systemfields import ModelPIDField
 from .names import record_type
 
-Name = record_type.record_cls
+
+class Name(record_type.record_cls):
+    """Name API class."""
+
+    pid = ModelPIDField()
```
slint, line 18: Comment: as discussed IRL, this can be avoided via changing the RecordTypeFactory parameters (still to be figured out how).

invenio_vocabularies/records/pidprovider.py
```diff
@@ -102,6 +102,27 @@ def create(cls, object_type=None, object_uuid=None, record=None, **kwargs):
...
+    field = "identifiers"
+    scheme = "orcid"
+
+    @classmethod
+    def create(cls, record):
+        """Extract the PID value."""
+        for id_ in record.get(cls.field):  # TODO: implement recursion?
+            if id_["scheme"] == cls.scheme:
+                return id_["identifier"]
+        return None
```
slint, line 123: Doubt: I'm a bit hesitant regarding this implicit "coupling" between the `identifiers` (or any) metadata field, and the systemfield itself. E.g. imagine one creates a record without including in `identifiers` a value with the required scheme, what happens to the `record.pid` value? I guess it's `None` which is an issue. 

In theory, the validation would need to happen on the service schema either to validate either:

- that an `identifiers` entry exists with `scheme: orcid`
- a `pid` value has been passed (and possibly validate that the `scheme` is correct?)

slint (commented):
LGTM, some comments on validation

invenio_vocabularies/records/pidprovider.py
ppanero: Yes, this is the drawback. I think we can do it in two steps:
1. _pre load_: if there is no pid, and take it from a configured field `identifiers`
2. _validate_: if no pid (extracted before), then fail (`pid = ...(required=True)`)

invenio_vocabularies/contrib/names/services.py
```diff
@@ -20,6 +20,8 @@
 class NamesService(record_type.service_cls):
     """Name service."""
 
+    # TODO: this method should be extracted to a common Mixin
```
ppanero, line 23: opened issue #162

ppanero:
closing PoC, will be added in #147

PR #163, opened 2022-03-15
Awards vocabulary

alejandromumo:
This PR fixes https://github.com/inveniosoftware/invenio-vocabularies/issues/158

### DIFF
--- a/invenio_vocabularies/alembic/e1146238edd3_create_awards_table.py
+++ b/invenio_vocabularies/alembic/e1146238edd3_create_awards_table.py
@@ -5,7 +5,7 @@
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
-"""Create funders vocabulary table."""
+"""Create awards table."""
 
 import sqlalchemy as sa
 import sqlalchemy_utils
@@ -13,16 +13,16 @@ from alembic import op
 from sqlalchemy.dialects import mysql, postgresql
 
 # revision identifiers, used by Alembic.
-revision = '676dd587542d'
-down_revision = '17c703ce1eb7'
+revision = 'e1146238edd3'
+down_revision = '676dd587542d'
 branch_labels = ()
 depends_on = None
 
 
 def upgrade():
-    """Upgrade database."""
+    """Downgrade database."""
     op.create_table(
-        'funder_metadata',
+        'award_metadata',
         sa.Column(
             'created',
             sa.DateTime().with_variant(mysql.DATETIME(fsp=6), 'mysql'),
@@ -50,13 +50,11 @@ def upgrade():
         ),
         sa.Column('version_id', sa.Integer(), nullable=False),
         sa.Column('pid', sa.String(), nullable=True),
-        sa.PrimaryKeyConstraint('id', name=op.f('pk_funder_metadata')),
-        sa.UniqueConstraint('pid', name=op.f('uq_funder_metadata_pid'))
+        sa.PrimaryKeyConstraint('id', name=op.f('pk_award_metadata')),
+        sa.UniqueConstraint('pid', name=op.f('uq_award_metadata_pid'))
     )
-    # ### end Alembic commands ###
 
 
 def downgrade():
-    """Downgrade database."""
-    op.drop_table('funder_metadata')
-    # ### end Alembic commands ###
+    """Upgrade database."""
+    op.drop_table('award_metadata')
--- a/invenio_vocabularies/config.py
+++ b/invenio_vocabularies/config.py
@@ -62,6 +62,17 @@ VOCABULARIES_FUNDER_SCHEMES = {
 VOCABULARIES_FUNDER_DOI_PREFIX = "10.13039"
 """DOI prefix for the identifier formed with the FundRef id."""
 
+VOCABULARIES_AWARD_SCHEMES = {
+    "url": {
+        "label": _("URL"),
+        "validator": idutils.is_url
+    },
+    "doi": {
+        "label": _("DOI"),
+        "validator": idutils.is_doi
+    },
+}
+"""Awards allowed identifier schemes."""
 
 VOCABULARIES_NAMES_SCHEMES = {
     "orcid": {
@@ -70,10 +81,10 @@ VOCABULARIES_NAMES_SCHEMES = {
         "datacite": "ORCID"
     },
     "isni": {
-         "label": _("ISNI"),
-         "validator": idutils.is_isni,
-         "datacite": "ISNI"
-     },
+        "label": _("ISNI"),
+        "validator": idutils.is_isni,
+        "datacite": "ISNI"
+    },
     "gnd": {
         "label": _("GND"),
         "validator": idutils.is_gnd,
--- a/invenio_vocabularies/contrib/awards/__init__.py
+++ b/invenio_vocabularies/contrib/awards/__init__.py
@@ -1,9 +1,19 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2021-2022 CERN.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
 """Awards vocabulary."""
+
+from .resources import AwardsResource, AwardsResourceConfig
+from .services import AwardsService, AwardsServiceConfig
+
+__all__ = [
+    "AwardsResource",
+    "AwardsResourceConfig",
+    "AwardsService",
+    "AwardsServiceConfig",
+]
--- a/invenio_vocabularies/contrib/awards/api.py
+++ b/invenio_vocabularies/contrib/awards/api.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary awards."""
+
+from .awards import record_type
+
+Award = record_type.record_cls
--- a/invenio_vocabularies/contrib/awards/awards.py
+++ b/invenio_vocabularies/contrib/awards/awards.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary awards."""
+
+from invenio_db import db
+from invenio_records.dumpers import ElasticsearchDumper
+from invenio_records.systemfields import RelationsField
+from invenio_records_resources.factories.factory import RecordTypeFactory
+from invenio_records_resources.records.systemfields import ModelPIDField, \
+    PIDListRelation
+
+from ...services.permissions import PermissionPolicy
+from ..funders.api import Funder
+from .config import AwardsSearchOptions, service_components
+from .schema import AwardSchema
+
+award_relations = RelationsField(
+    funders=PIDListRelation(
+        'funders',
+        attrs=['id', 'name'],
+        pid_field=Funder.pid,
+        cache_key='funders',
+    )
+)
+
+record_type = RecordTypeFactory(
+    "Award",
+    # Data layer
+    pid_field_cls=ModelPIDField,
+    pid_field_kwargs={
+        "model_field_name": "pid",
+    },
+    model_cls_attrs={
+        # cannot set to nullable=False because it would fail at
+        # service level when create({}), see records-resources.
+        "pid": db.Column(db.String, unique=True),
+    },
+    record_dumper=ElasticsearchDumper(
+        model_fields={'pid': ('pid', str)}
+    ),
+    schema_version="1.0.0",
+    schema_path="local://awards/award-v1.0.0.json",
+    record_relations=award_relations,
+    # Service layer
+    service_schema=AwardSchema,
+    search_options=AwardsSearchOptions,
+    service_components=service_components,
+    permission_policy_cls=PermissionPolicy,
+    # Resource layer
+    endpoint_route='/awards',
+)
--- a/invenio_vocabularies/contrib/awards/awards.py
+++ b/invenio_vocabularies/contrib/awards/awards.py
@@ -13,7 +13,7 @@ from invenio_records.dumpers import ElasticsearchDumper
 from invenio_records.systemfields import RelationsField
 from invenio_records_resources.factories.factory import RecordTypeFactory
 from invenio_records_resources.records.systemfields import ModelPIDField, \
-    PIDListRelation
+    PIDRelation
 
 from ...services.permissions import PermissionPolicy
 from ..funders.api import Funder
@@ -21,11 +21,11 @@ from .config import AwardsSearchOptions, service_components
 from .schema import AwardSchema
 
 award_relations = RelationsField(
-    funders=PIDListRelation(
-        'funders',
+    funders=PIDRelation(
+        'funder',
         attrs=['id', 'name'],
         pid_field=Funder.pid,
-        cache_key='funders',
+        cache_key='funder',
     )
 )
 
--- a/invenio_vocabularies/contrib/awards/config.py
+++ b/invenio_vocabularies/contrib/awards/config.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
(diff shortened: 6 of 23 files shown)


invenio_vocabularies/alembic/e1146238edd3_create_awards_table.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""Create awards table."""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = 'e1146238edd3'
+down_revision = '3d362de4926a'
```
ppanero, line 16: Should the revision be `8ff82dfb0be8` (vocabularies branch)? In theory `awards` does not need funders (i.e. in case all funders are **not** from a vocabulary). In that case, it should just be dependent on the start of vocabularies.

invenio_vocabularies/config.py
```diff
@@ -42,8 +42,23 @@
...
+}
 """Affiliations allowed identifier schemes."""
 
+VOCABULARIES_FUNDER_SCHEMES = {
+    **VOCABULARIES_IDENTIFIER_SCHEMES,
+}
+"""Funders allowed identifier schemes."""
+
+VOCABULARIES_AWARD_SCHEMES = {
+    **VOCABULARIES_IDENTIFIER_SCHEMES,
```
ppanero, line 58: minor: are we sure these are the same schemes? I see rdm-records had no list for starters... but I think for example "doi" is missing? @slint

invenio_vocabularies/contrib/awards/api.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary awards."""
+
```
ppanero, line 10:
```suggestion
```

invenio_vocabularies/contrib/awards/awards.py
```diff
@@ -0,0 +1,36 @@
...
+
+from invenio_records_resources.factories.factory import RecordTypeFactory
+
+from ...records.pidprovider import PIDProviderFactory
+from ...records.systemfields import BaseVocabularyPIDFieldContext
+from ...services.permissions import PermissionPolicy
+from .config import AwardsSearchOptions, service_components
+from .schema import AwardSchema
+
+record_type = RecordTypeFactory(
```
ppanero, line 19: major: it's missing the relation ("_programmatic fk_"), therefore  it is not being linked to a funder. See [names](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/names/names.py#L23) as an example. This also means that when doing an award test the funder needs to be in the DB (i.e. use both funder and award fixture -  see names)

invenio_vocabularies/contrib/awards/mappings/v7/awards/award-v1.0.0.json
```diff
@@ -0,0 +1,84 @@
...
+            "type": "keyword"
+          },
+          "id": {
+            "type": "keyword"
+          },
+          "name": {
+            "type": "keyword",
+            "fields": {
+              "suggest": {
+                "type": "search_as_you_type"
```
ppanero, line 76: minor: I'm not sure it is needed. It's what we discussed, will depend on the UI cases:
- Users look for funders,
- Then for an award inside a funder (in which case, with a "keyword" name we can just filter).

invenio_vocabularies/contrib/awards/models.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary awards."""
```
ppanero, line 9:
```suggestion
"""Awards vocabulary models."""
```

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -9,26 +9,53 @@
...
+    number = SanitizedUnicode()
+    title = i18n_strings
+
+    @validates_schema
+    def validate_data(self, data, **kwargs):
+        """Validate either id or number/title are present."""
+        id_ = data.get("id")
+        number = data.get("number")
+        title = data.get("title")
+        if (id_) or (number and title):
```
ppanero, line 52: This is more than okay, just some options... this reaaaally depends on your coding style (and there is no correct answer here).
```suggestion
        if id_ or (number and title):
```

This would also work, and since the values are used **only** for a condition:
```python
is_id = data.get("id")
is_custom = data.get("number") and data.get("title")
if not is_id and not is_custom:
    raise
```

In any case, I would invert the conditions to avoid the `pass + else` and just have one `if`

tests/contrib/awards/test_awards_api.py
```diff
@@ -0,0 +1,101 @@
...
+    """Award schema validation."""
+    # valid data
+    awa = example_award
+
+    assert awa.schema == "local://awards/award-v1.0.0.json"
+    assert awa.pid
+    assert awa.id
+
+    # invalid data
+    examples = [
```
ppanero, line 58: minor: maybe we should ask @slint @ctrl-alt-soneca for some real life example of awards so we can abstract data properly. Just wondering if someone in the future sees this tests the data looks more like affiliations than actual awards (could be ok for funders tho...)

tests/contrib/awards/test_awards_resource.py
```diff
@@ -0,0 +1,153 @@
...
+    assert res.json["hits"]["total"] == 1
+    assert res.json["sortBy"] == "bestmatch"
+
+
+def _create_awards(service, identity):
+    """Create dummy awards with similar ids/numbers/titles."""
+    awards = [
+        {
+            "title": {
+                "en": "European Organization for Nuclear Research",
```
ppanero, line 96: minor: same as above, we should find another example. Otherwise, is not representative of the use case and we are "roguing" the test.

invenio_vocabularies/contrib/awards/schema.py
alejandromumo: Ended up with something similar, without changing the variables instantiation:

```python
 if not id_ and not (number and title):
    # raise exception      
```

invenio_vocabularies/contrib/awards/awards.py
```diff
@@ -0,0 +1,49 @@
...
+from invenio_records_resources.records.systemfields import PIDListRelation
+
+from ...records.pidprovider import PIDProviderFactory
+from ...records.systemfields import BaseVocabularyPIDFieldContext
+from ...services.permissions import PermissionPolicy
+from ..funders.api import Funder
+from .config import AwardsSearchOptions, service_components
+from .schema import AwardSchema
+
+name_relations = RelationsField(
```
ppanero, line 22:
```suggestion
awards_relations = RelationsField(
```

invenio_vocabularies/contrib/awards/awards.py
```diff
@@ -0,0 +1,49 @@
...
+    "Award",
+    # Data layer
+    pid_field_kwargs={
+        "create": False,
+        "provider": PIDProviderFactory.create(pid_type='awa'),
+        "context_cls": BaseVocabularyPIDFieldContext,
+    },
+    schema_version="1.0.0",
+    schema_path="local://awards/award-v1.0.0.json",
+    record_relations=name_relations,
```
ppanero, line 41:
```suggestion
    record_relations=awards_relations,
```

invenio_vocabularies/contrib/awards/mappings/v7/awards/award-v1.0.0.json
```diff
@@ -0,0 +1,79 @@
...
+        "type": "object",
+        "properties": {
+          "@v": {
+            "type": "keyword"
+          },
+          "id": {
+            "type": "keyword"
+          },
+          "name": {
+            "type": "keyword"
```
ppanero, line 73: keyword or text?

alejandromumo: I think text would be the correct one. Nonetheless, `names` vocabulary also defines `affiliations` names as `keyword`. We might want to change that as well?

ppanero: Yes, but not in the same PR. Could you open a quick issue (no need for a big description) and PR, and add the issue to the board?

invenio_vocabularies/contrib/awards/__init__.py
```diff
@@ -1,9 +1,19 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2022 CERN.
```
ppanero, line 3:
```suggestion
# Copyright (C) 2021-2022 CERN.
```

ppanero, line 3:
```suggestion
# Copyright (C) 2021-2022 CERN.
```

invenio_vocabularies/contrib/awards/awards.py
```diff
@@ -0,0 +1,64 @@
...
+from invenio_records_resources.records.systemfields import ModelPIDField, \
+    PIDListRelation
+from ...services.permissions import PermissionPolicy
+from sqlalchemy_utils.types import ChoiceType
+from ..funders.api import Funder
+from .config import AwardsSearchOptions, service_components
+from .schema import AwardSchema
+
+award_relations = RelationsField(
+    funders=PIDListRelation(
```
ppanero, line 25: comment: Im not a 100% sure this is gonna work with the new `ModelPIDField`, since the `pid_field` is now a `PersistentIdentifierWrapper` and it might be expected to have some funcs that are not available. TBD.

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -9,26 +9,90 @@
...
+    identifiers = IdentifierSet(fields.Nested(
+        partial(
+            IdentifierSchema,
+            allowed_schemes=award_schemes,
+            identifier_required=False
+        )
+    ))
     number = SanitizedUnicode(
         required=True,
         validate=validate.Length(min=1, error=_('Number cannot be blank.'))
```
ppanero, line 34: minor: is this needed? does the required check for "not blank"? I guess it just checks its presence?

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -9,26 +9,90 @@
...
+        for identifier in identifiers:
+            # FIXME: make scheme configurable
+            if identifier["scheme"] == "ror":
+                data["pid"] = identifier["identifier"]
+                break
+
+        return data
+
+    @post_load
+    def kepp_pid_on_create(self, data, **kwargs):
```
ppanero, line 57: this typo was mine 😂 

```suggestion
    def keep_pid_on_create(self, data, **kwargs):
```

tests/contrib/awards/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+            "en": "European Organization for Nuclear Research",
+            "fr": "Organisation européenne pour la recherche nucléaire"
+        },
+        "country": "CH"
+    }
+    # at API level 'pid' is passed as an arg
+    fun = Funder.create(api_funder, pid="01ggx4157")
+    fun.commit()
+    db.session.commit()
+    return fun
```
ppanero, line 91: minor: won't we have the same problem here with the soft deletion? (i.e. we have to yield + hard delete?)

tests/contrib/awards/test_awards_api.py
```diff
@@ -0,0 +1,95 @@
...
+@pytest.fixture()
+def search_get():
+    """Get a document from an index."""
+    return partial(
+        current_search_client.get, Award.index._name, doc_type="_doc"
+    )
+
+
+@pytest.fixture()
+def example_award(db, award_full_data, example_funder):
```
ppanero, line 31: moderate: duplicated from conftest? are both needed?

invenio_vocabularies/contrib/awards/mappings/v7/awards/award-v1.0.0.json
alejandromumo: Created it here https://github.com/inveniosoftware/invenio-vocabularies/issues/166

invenio_vocabularies/config.py
alejandromumo: As discussed with @slint , `doi` and `url` are valid schemes for an award for now.

tests/contrib/awards/test_awards_api.py
alejandromumo: I am afraid that, for now, both are needed indeed similarly to funders. 
At this level, we want to create an award programmatically, while in other tests we want to create an award using the service's proxy.

tests/contrib/awards/conftest.py
ppanero: The problem wasn't appearing because the fixture was not used more than once in the same module 😅 I added the deletion.

invenio_vocabularies/contrib/awards/awards.py
ppanero: Updated the relation config, and tests for it, but it is working 🚀

invenio_vocabularies/contrib/awards/jsonschemas/awards/award-v1.0.0.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "identifiers": {
+      "description": "Alternate identifiers for the award.",
+      "type": "array",
+      "items": {
+        "$ref": "local://definitions-v1.0.0.json#/identifiers_with_scheme"
+      },
+      "uniqueItems": true
+    },
+    "number": {
```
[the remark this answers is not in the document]
ppanero: Added. check https://github.com/inveniosoftware/invenio-vocabularies/pull/163/commits/75078a996c517ebf90b6a213ee5448e98082b030

invenio_vocabularies/contrib/awards/config.py
```diff
@@ -0,0 +1,43 @@
...
+    """Search options."""
+
+    suggest_parser_cls = SuggestQueryParser.factory(
+        fields=[
+            'title.*^50',
+            'title.*._2gram',
+            'title.*._3gram',
+            'number^10'
+        ],
+    )
```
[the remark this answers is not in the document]
slint: It could, but I feel with a facet it's a bit more "streamlined", i.e. no need to string-format to build a query and potentially have to escape characters or whatever... Could also be just mental leftovers from Zenodo :)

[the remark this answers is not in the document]
ppanero: you are right, we would need to reinvent the wheel for formatting. I only had the filtering case in mind.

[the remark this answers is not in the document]
ppanero: @slint This PR is ready, faceting configuration happens at RDM-Records level (see [example](https://github.com/inveniosoftware/invenio-rdm-records/blob/aaf8df982e9099b12b38d3374519669c9ca24f68/invenio_rdm_records/services/facets.py#L46)) and I think we can do with the `VocabularyLabels`, which will show the i18n title of the funder/award. And are configured [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/cc8514638bbd052ae211772c8ca542fa01f5c4ec/invenio_rdm_records/config.py#L308)... I'm not 100% sure this fulfills our use case though.

What we want is a facet over the funder when querying awards, right? I think in RDM Records all faceting is configured **only** for the main search. So we might need to do some extra work. Treat in a different PR?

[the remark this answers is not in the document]
slint: This is actually for filtering over `/api/awards`, and more specifically for the suggester (though I think it wouldn't matter with the way we build queries?). Imagine the following:

User selects the funder "European Commision" (with `pid = 00k4n6c32`) from the modal's dropdown. Afterwards, there's the first query on `/api/awards?funder=00k4n6c32&suggest=` (no search query was typed by the user), which returns only awards from the selected funder. Then if the user types e.g. an acronym in the search box (e.g. `Si-POP`), we do the query `/api/awards?funder=00k4n6c32&suggest=Si-POP`.

Then, there are things to take into account on how we search published records, but can be discussed separately. One example is "I want all records funded by the EC", which translates to something like `/api/records?q=funding.funder.id:00k4n6c32`.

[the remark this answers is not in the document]
ppanero: Okay, I got it now (thanks for the detailed explanation), it would be done in vocabularies then. However, since accounts got fixed... maybe we can start merging and treating in separate prs so manuel does not need to install from many masters 😅

PR #167, opened 2022-03-24
Awards datastreams

mitsosf:


invenio_vocabularies/config.py
```diff
@@ -74,6 +74,33 @@
 }
 """Awards allowed identifier schemes."""
 
+VOCABULARIES_AWARD_OPENAIRE_FUNDERS = {
```
ppanero, line 77: comment:

at some point we will have an awards dump in yaml, made based on some filtering and we will load it to Zenodo. Why don't we add this as a property e.g. "openaire_id", which can be resolved from DB? This would mean that new funders/updates just need to be processed as any other vocabulary update.


p.s. sorry for "eavesdropping" 😇

invenio_vocabularies/datastreams/readers.py
```diff
@@ -159,6 +160,32 @@ def _iter(self, fp, *args, **kwargs):
...
+            except JSONDecodeError as err:
+                raise ReaderError(
+                    f"Cannot decode JSON line {fp.name}:{idx}: {str(err)}"
+                )
+
+
+class GzipReader(BaseReader):
+    """Gzip reader."""
+
+    def _iter(self, fp, *args, **kwargs):
```
ppanero, line 184: comment: I think this should be part of the `read`, otherwise the piping might not work. e.g.

1. Tar reader yields file paths
2. They are received by this reader, who's read function would `.open()` since it comes in the `item` parameter. Then it would try to `re-open` in the `_iter` steps... 

The same happened to the tar and zip reader whose functionality got splited.

p.s. sorry for eavesdropping again, just got a notification from GH 😅

slint: In the end, this one worked out since `gzip.open(...)` accepts file pointers as well (which is done in `BaseReader.read()`)

invenio_vocabularies/datastreams/writers.py
```diff
@@ -86,6 +87,9 @@ def write(self, stream_entry, *args, **kwargs):
 
         except ValidationError as err:
             raise WriterError([{"ValidationError": err.messages}])
+        except InvalidRelationValue as err:
+            # TODO: Check if we can get the error message easier
```
ppanero, line 91: question: I guess `err.messages` does not work?

ppanero (approved):

invenio_vocabularies/config.py
```diff
@@ -74,6 +74,36 @@
...
+    'fct_________': '00snfqn58',
+    'fwf_________': '013tf3c58',
+    'irb_hr______': '03n51vw80',
+    'mestd_______': '01znas443',
+    'nhmrc_______': '011kf5r70',
+    'nih_________': '01cwqze88',
+    'nserc_______': '01h531d29',
+    'nsf_________': '021nxhr62',
+    'nwo_________': '04jsz6e67',
+    'rcuk________': '10.13039/501100000690',
```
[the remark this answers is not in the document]
slint: Yeah, unfortunately, this is a "deprecated" funder that is not part of the ROR dump (but which exists in the OpenAIRE awards dump). We'll have to make a "one-off" import with the correct metadata probably to supplement.

PR #169, opened 2022-03-25
Awards facets

ppanero:
- requires #163 
- closes https://github.com/inveniosoftware/invenio-vocabularies/issues/168 

## Changes

- Enable customization of service used to get the vocabularies when calculating the facet labels.
- Enable customization of the label string (i.e. not just `title`).
- Implements a `FundersLabels`, with the corresponding changes in `service.read_many`.

### DIFF
--- a/invenio_vocabularies/contrib/funders/facets.py
+++ b/invenio_vocabularies/contrib/funders/facets.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Vocabulary awards facets and labels."""
+
+from ...services.facets import VocabularyLabels, lazy_get_label
+
+
+class FundersLabels(VocabularyLabels):
+    """Fetching of vocabulary labels for facets."""
+
+    def __init__(self, vocabulary, cache=False, service_name=None):
+        """Initialize the labels."""
+        super().__init__(
+            vocabulary, cache, service_name="funders-service", id_field="pid"
+        )
+        self.fields = ["pid", "title", "country"]  # not configurable
+
+    def _vocab_to_label(self, vocab):
+        """Returns the label string for a vocabulary entry."""
+        return f"{lazy_get_label(vocab['title'])} ({vocab['country']})"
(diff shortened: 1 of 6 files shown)


invenio_vocabularies/contrib/funders/facets.py
```diff
@@ -0,0 +1,26 @@
...
+from ...services.facets import VocabularyLabels, lazy_get_label
+
+
+class FundersLabels(VocabularyLabels):
+    """Fetching of vocabulary labels for facets."""
+
+    def __init__(self, vocabulary, cache=False, service_name=None):
+        """Initialize the labels."""
+        super().__init__(
+            vocabulary, cache, service_name="funders-service", id_field="pid"
```
ppanero, line 20: The `service_name` is needed to avoid using always the generic vocbulary service. However, this means that we need to standarize the way we register services. Currently we are using:
- <vocabulary_name>-service when in `invenio-vocabularies`. For example: `awards-service`, `names-service`.
- rdm-<vocabulary_name> when in `invenio-rdm-records`. For example: `rdm-awards`, `rdm-names`.

At the moment this means that labelling will fail when integrated.

slint: I agree... Would it make sense to only configure the facets in `rdm-records` maybe? We can still test in this module (with the "mock" `awards-service` configured).

tests/contrib/awards/test_awards_resource.py
```diff
@@ -70,6 +70,13 @@ def test_awards_search(client, example_award, h, prefix):
     assert res.status_code == 200
     assert res.json["hits"]["total"] == 1
     assert res.json["sortBy"] == "newest"
+    assert res.json["aggregations"]["funders"]
+
+    funders_agg = res.json["aggregations"]["funders"]["buckets"][0]
+    assert funders_agg["key"] == "01ggx4157"
+    assert funders_agg["doc_count"] == 1
+    assert funders_agg["label"] == \
+        "European Organization for Nuclear Research (CH)"
```
slint, line 79: Can we add a test using both the facet and suggest? This will be closer to the "real-life" usage also.

slint (approved):

tests/contrib/awards/test_awards_resource.py
ppanero: Done, split in two:
- test_suggest_sort (previous test).
- test_faceted_suggest, with a query that should return two but only returns one because it is filtered by funder.

invenio_vocabularies/contrib/funders/facets.py
ppanero: This actually make sense, we do register the facets in rdm-records for bib records (instead of drafts- or records- resources).

PR #175, opened 2022-03-31
tests: remove python 3.6 support and update workflow file

ppanero:


setup.py
```diff
@@ -60,8 +60,7 @@
 ]
 
 install_requires = [
-    "invenio-records-resources>=0.19.4,<0.20.0",
-    "invenio-i18n>=1.3.1",
```
ppanero, line 64: comes from invenio-records-resources, duplicating causes version conflicts and we can actually rely on what comes from records-resources

PR #199, opened 2022-06-10
serializar: add localization serializer

jrcastro2:
* Add serializer to awards and funder
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1518


invenio_vocabularies/resources/serializer.py
```diff
@@ -45,6 +45,6 @@ class VocabularyL10NListSchema(Schema):
 
     def get_hits(self, obj_list):
         """Apply hits transformation."""
-        schema = self.context["schema_cls"]()
+        schema = self.context["object_schema_cls"]()
```
zzacharo, line 48: Shouldn't all these schemas inherit from the base schema in flask-resources?

jrcastro2: I guess we could, we would also dump the non-existing aggregations but I don't think that's a problem

invenio_vocabularies/resources/serializer.py
```diff
@@ -38,13 +39,5 @@ class VocabularyL10NItemSchema(Schema):
     tags = fields.List(fields.Str(), dump_only=True)
 
 
-class VocabularyL10NListSchema(Schema):
+class VocabularyL10NListSchema(BaseListSchema):
```
zzacharo, line 42: I guess you can remove that class and use directly the `BaseListSchema` class here?

jrcastro2: Yes, we can, just thought it would be easier to read the code when we "group" vocabulary related class, instead of calling the parent class directly. I will change it

zzacharo (approved):

PR #203, opened 2022-07-15
datastreams: contrib service writers state the service name

ppanero:
**Needs tests fixing**

### DIFF
--- a/invenio_vocabularies/datastreams/writers.py
+++ b/invenio_vocabularies/datastreams/writers.py
@@ -39,7 +39,7 @@ class BaseWriter(ABC):
 class ServiceWriter(BaseWriter):
     """Writes the entries to an RDM instance using a Service object."""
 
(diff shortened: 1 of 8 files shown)


invenio_vocabularies/datastreams/writers.py
```diff
@@ -39,7 +39,7 @@ def write(self, stream_entry, *args, **kwargs):
 class ServiceWriter(BaseWriter):
     """Writes the entries to an RDM instance using a Service object."""
 
-    def __init__(self, service_or_name, identity, *args, update=False, **kwargs):
+    def __init__(self, service_or_name, *args, identity=None, update=False, **kwargs):
```
ppanero, line 42: This is backwards incompatible. However, nobody should be using it since before we were just ignoring the `identity`. Therefore can go in a minor/patch release.

PR #225, opened 2022-09-20
custom fields: use underscore to split namespace from name in relatio…

### LINKED ISSUE #224: replace `:` by `_` in namespaces for cf relations
https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/records/systemfields/relations.py#L38

ppanero:
- closes #224

### DIFF
--- a/invenio_vocabularies/records/systemfields/relations.py
+++ b/invenio_vocabularies/records/systemfields/relations.py
@@ -35,8 +35,9 @@ class CustomFieldsRelation(RelationsField):
         relations = {}
         for cf in custom_fields:
             if getattr(cf, "relation_cls", None):
-                relations[cf.name] = cf.relation_cls(
-                    f"custom_fields.{cf.name}",
+                rel_name = cf.name.replace(":", "_")  # for namespaces
+                relations[rel_name] = cf.relation_cls(
+                    f"custom_fields.{cf.name}",  # needs : due to ES field name
                     keys=cf.field_keys,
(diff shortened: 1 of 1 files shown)


invenio_vocabularies/records/systemfields/relations.py
```diff
@@ -35,8 +35,9 @@ def _fields(self):
         relations = {}
         for cf in custom_fields:
             if getattr(cf, "relation_cls", None):
-                relations[cf.name] = cf.relation_cls(
-                    f"custom_fields.{cf.name}",
+                rel_name = cf.name.replace(":", "_")  # for namespaces
+                relations[rel_name] = cf.relation_cls(
```
slint, line 39: I guess this is for being able to `.`-access fields without having to use `getattr`? Not sure how often it's needed but it's probably good for debugging in any case.

slint (approved):
LGTM, just a comment on the usage of this.

invenio_vocabularies/records/systemfields/relations.py
ppanero: Yes, is it would be slightly inconsistent naming wise (i.e. json attr vs programatic attr). But with the `:` we could not `.` access the field.

zzacharo: I had ended up to the same thing in the past but I kept it like that so we are consistent. Otherwise, it will create confusion....

ppanero:
Not clear if we want to do this atm, it might be hard to talk to the users. Is just a programatic issue.

PR #231, opened 2022-11-02
Read-only mode

max-moser:
Requires https://github.com/inveniosoftware/invenio-records-permissions/pull/86


max-moser:
The tests here go via the services rather than the resources/API because no write-operation endpoints are exposed.

utnapischtim:
closed due to inactivity

PR #243, opened 2023-01-06
facets: handle non-existent vocabulary type

ppanero:
- requires https://github.com/inveniosoftware/invenio-records-resources/pull/435
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1952

In invenio-vocabularies we have `read`, `read_all` and `read_many`. When searching for generic vocabularies, we give two parameters vocabulary `type` and `ids` (e.g. _resourcetype_ and _publication_). In the case that publication does not exist the result would be:
- `read` --> `NoResultFound` (404) 
- `read_many` and `read_all` --> empty search result (i.e. `hits: []`)

However, all three methods assume that the vocabulary type (e.g. _resourcetype_) exists in db (we are querying `.one()` with no try/except). Then, it make the search page with a 404 because a `NoResultFound` is risen. It leads me to:
- `read` --> `NoResultFound` (404) . This is still correct, however, should we expose in the error if what does exist is the type instead of the actual id?
- `read_many` and `read_all` --> **Here is the important part**:
    - Imitating the empty search result (i.e. `hits: []`) would be the desired output for internal usage (e.g. facets). However, it is semantically incorrect, since it's not empty but rather doesn't exist.
    - 404, would need to handle this internally when labelling vocabularies <-- I'm leaning towards this solution, but would mean silent failure

### DIFF
--- a/invenio_vocabularies/services/facets.py
+++ b/invenio_vocabularies/services/facets.py
@@ -13,6 +13,7 @@ from invenio_i18n.ext import current_i18n
 from invenio_records_resources.proxies import current_service_registry
 from marshmallow_utils.fields.babel import gettext_from_dict
 from speaklater import make_lazy_string
+from sqlalchemy.exc import NoResultFound
 
 from ..proxies import current_service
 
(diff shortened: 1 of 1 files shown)


invenio_vocabularies/services/facets.py
```diff
@@ -53,17 +54,22 @@ def _vocab_to_label(self, vocab):
...
+                )
+            else:
+                vocabs = self.service.read_all(
+                    identity, type=self.vocabulary, fields=self.fields
+                )
+            vocab_list = list(vocabs.hits)  # the service returns a generator
+        except NoResultFound:
+            # in the case the vocabulary type does not exist
+            # facets should not fail but be empty
+            vocab_list = []
```
ppanero, line 70: **alternative**: create a custom exception, that then would be raised in the except and captured as a 500.

ntarocco: Question: this is called if the facet is defined, and it should work...
If I don't define a facet on this vocabulary, then it should not be called (so it should not fail).

If I configure the facet, I would probably prefer to have a clear exception. Instead, I see an empty facet and I won't understand why?
Let me know if I have maybe misunderstood this...

ppanero: What you are saying is correct. I still think a full blown 500 is an overkill for something that has a small-medium impact on the user. But since we are not using logging consistently (i.e. to alert sysadmins), I will create a new error (e.g. `FacetNotFound` and raise it)

ppanero:
How it looks:

Screenshot 2023-01-06 at 16 37 59

invenio_vocabularies/services/facets.py
ppanero: Implemented as discussed in Discord

jrcastro2 (approved):

ppanero:
test fail due to missing requirement https://github.com/inveniosoftware/invenio-records-resources/pull/435

ntarocco (approved):
LGTM! Thanks!

PR #247, opened 2023-02-20
mappings: add dynamic template for i18n titles and descriptions

### LINKED ISSUE #246: 'description' mapping is declared as an object with dynamic keys without any template
Description mapping is declared as an [object](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/records/mappings/os-v2/vocabularies/vocabulary-v1.0.0.json#L73) with dynamic keys, without specifying any template at all (e.g. `i18n` keys).

It should have a template (e.g. [`title`](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/records/mappings/os-v2/vocabularies/vocabulary-v1.0.0.json#L63)) to validate `i18n` keys.

ppanero:
- closes #246

### DIFF
--- a/invenio_vocabularies/records/mappings/os-v1/vocabularies/vocabulary-v1.0.0.json
+++ b/invenio_vocabularies/records/mappings/os-v1/vocabularies/vocabulary-v1.0.0.json
@@ -1,5 +1,24 @@
 {
   "mappings": {
+    "dynamic_templates": [
+      {
+        "i18n_title": {
+          "path_match": "title.*",
+          "match_mapping_type": "string",
+          "mapping": {
+            "type": "search_as_you_type"
+          }
+        }
+      }, {
+        "i18n_description": {
+          "path_match": "description.*",
+          "match_mapping_type": "string",
+          "mapping": {
+            "type": "text"
+          }
+        }
+      }
+    ],
     "dynamic": "strict",
     "properties": {
       "$schema": {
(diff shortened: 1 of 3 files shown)


kpsherva (approved):

ppanero:
@slint unfortunately there is nothing like that at the moment. I could add it, but that is extra work that I'm not sure it fits on the current sprint. At the moment we are testing this indirectly, if you pass a normal string here https://github.com/inveniosoftware/invenio-vocabularies/blob/25ba54bbaee5fd89794067b2c34c6bea20e7fc16/tests/conftest.py#L134 it should fail. I could may do that test, but it feels ad-hoc (note that we have this type of dynamic templates also in rdm-records)

slint:
Fine with me, was curious because ES mapping errors/mismatches are usually going under the radar, but if there's code that works with the functionality it should be good :)

ppanero:
@slint sample run:

**Normal run**

```python
@pytest.fixture(scope="module")
def example_data():
    """Example data for records."""
    return {
    ...
        },
        "description": {
            "en": "Publications",
            "da": "Publikationer",
        },
    ...
    }

tests/resources/test_resources_l10n.py::test_get PASSED
```

**Not a dict**, this would fail even without dynamic template. The check is dict vs no dict.
```python
@pytest.fixture(scope="module")
def example_data():
    """Example data for records."""
    return {
        ...
        "description": "Publications",
        ...
    }

ERROR tests/resources/test_resources_l10n.py::test_get - marshmallow.exceptions.ValidationError: {'description': ['Not a valid mapping type.']}
```

**Not a string**, this is enforced by the template. However, we can only check that is a string we cannot check that is indexed as `search_as_you_type`
```python
@pytest.fixture(scope="module")
def example_data():
    """Example data for records."""
    return {
        ...
        "description": {
            "en": 123,
        },
        ...
    }
    
marshmallow.exceptions.ValidationError: {'description': defaultdict(<class 'dict'>, {'en': {'value': ['Not a valid string.']}})}
```

PR #264, opened 2023-09-07
contrib: awards fixes

slint:
- config: update awards funders mapping
- awards: add "program" field


slint:
Included in https://github.com/inveniosoftware/invenio-vocabularies/pull/265

PR #271, opened 2023-09-18
facets: implement in-memory cache

ntarocco:
To be deployed and tested in an env.

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/facets.py
+++ b/invenio_vocabularies/contrib/affiliations/facets.py
@@ -14,14 +14,22 @@ from ...services.facets import VocabularyLabels, lazy_get_label
 class AffiliationsLabels(VocabularyLabels):
     """Fetching of vocabulary labels for facets."""
 
-    def __init__(self, vocabulary, cache=False, service_name=None):
-        """Initialize the labels."""
+    def __init__(self, vocabulary, cache=True, cache_ttl=3600, service_name=None):
+        """Initialize the labels.
+
+        :param vocabulary: the name of the vocabulary type.
+        :param cache: use simple process in-memory cache when True.
+        :param cache_ttl: cache expiration in seconds.
+        :param service_name: the name of the registered service to be used
+            when fetching values for the vocabulary.
+        """
         super().__init__(
             vocabulary,
-            cache,
+            cache=cache,
+            cache_ttl=cache_ttl,
             service_name="affiliations",
         )
-        self.fields = ["id", "title"]  # not configurable
+        self.fields = ("id", "title")  # not configurable
 
     def _vocab_to_label(self, vocab):
         """Returns the label string for a vocabulary entry."""
--- a/invenio_vocabularies/services/facets.py
+++ b/invenio_vocabularies/services/facets.py
@@ -8,6 +8,9 @@
 
 """Vocabulary facets."""
 
+import time
+from functools import lru_cache
+
 from flask_principal import AnonymousIdentity
 from invenio_i18n.ext import current_i18n
 from invenio_records_resources.proxies import current_service_registry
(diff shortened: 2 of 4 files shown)


ntarocco:
@ppanero if you have some time, can you please have a quick look if I missed anything?

invenio_vocabularies/contrib/affiliations/facets.py
```diff
@@ -14,14 +14,22 @@
 class AffiliationsLabels(VocabularyLabels):
     """Fetching of vocabulary labels for facets."""
 
-    def __init__(self, vocabulary, cache=False, service_name=None):
-        """Initialize the labels."""
+    def __init__(self, vocabulary, cache=True, cache_for=3600, service_name=None):
```
ppanero, line 17: ultra-minor: I thought it was cache "for this vocabulary" like a key, maybe `ttl` or `expiration`.

ppanero (approved):
Nice work! At first glance I do not see any issue. I'm wondering why it was done specifically on funcers and affiliations and not more generic:
- in VocabularyLabels
- a subclass of the above CacheVocabularyLabels (this would be a new class

I guess for now we only need to optimize those too? What about subjects (e.g. MeSH is very large).

invenio_vocabularies/contrib/affiliations/facets.py
ntarocco: good point, I will change it

invenio_vocabularies/services/facets.py
```diff
@@ -19,6 +22,49 @@
...
+    return list(vocabs.hits)  # the service returns a generator
+
+
+def get_ttl_hash(seconds=3600):
+    """Return the same value withing `seconds` time period. Default 1 hour."""
+    return round(time.time() / seconds)
+
+
+@lru_cache(maxsize=32)
+def cached_vocabs(service_name, type, ids, fields, ttl_hash=None):
```
slint, line 49: question/shleve: from what I understand this caches results for a service/vocab-type on the level of `ids`, i.e. the calls with ids `["dataset", "software", "figure"]` and `["software"]` are stored separately?

I guess this is to keep things simple, otherwise caching on individual IDs and reconstructing the result gets a bit complicated.

ntarocco: > question/shleve: from what I understand this caches results for a service/vocab-type on the level of `ids`, i.e. the calls with ids `["dataset", "software", "figure"]` and `["software"]` are stored separately?
> 
> I guess this is to keep things simple, otherwise caching on individual IDs and reconstructing the result gets a bit complicated.

Yes indeed. Would you prefer to cache individual ids?
The cache has been created having its usage in mind, where most of the time, a list of `ids` is passed. We could go for caching each `id` too, it might complexify things.
Do you see any particular advantage? The only advantage I can think of is when only one `id` is fetched. It will work anyway, but the first time you will have a `miss`.

slint: I think the list of IDs changes e.g. based on the search query you make, i.e. if I search records for `Python`, I will probably not get any `Text` resource type results in the facets.

Given that this is sort of a micro-caching strategy, on the level of individual uWSGI processes it should be fine for now (also it will get more complex to mangle the cached + fetched results). e.g. in Zenodo there's a big enough volume of records spread across all the different resource types, so the set of unique resource type IDs coming from the facet/aggregatio when you make a search query is likely to be the same.

ntarocco: Yes I agree. It was indeed done like this for simplicity and to measure if we really gain performance.
I can spend more time now to improve it, caching by id and handling the cache reconciliation of each id when multiple ids are fetched.

PR #277, opened 2023-09-28
facets: cache vocabularies by id

### LINKED ISSUE #275: Improve caching
The current vocabulary facet caching is suboptimal, because a set of given ids are cached together.
Ideally, we should cache each ID for better performance.

ntarocco:
* closes #275

### DIFF
--- a/invenio_vocabularies/services/facets.py
+++ b/invenio_vocabularies/services/facets.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021-2022 CERN.
+# Copyright (C) 2021-2023 CERN.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


invenio_vocabularies/services/facets.py
```diff
@@ -100,36 +94,36 @@ def _vocab_to_label(self, vocab):
...
+                    if id_ in ids:
+                        labels[id_] = self._vocab_to_label(vocab)
         except NoResultFound:
             raise FacetNotFoundError(self.vocabulary)
 
-        labels = {}
-        ids = set(ids)
-        seen = set()
-        for vocab in vocab_list:
-            # cannot loop over ids because vocab id is inside each item
```
ntarocco, line 126: I don't understand this comment `# cannot loop over ids because vocab id is inside each item`... is it valid?

PR #297, opened 2024-02-16
cli : Safely check writers 'args' on update

### LINKED ISSUE #292: datastreams: `vocabularies update` crash
Package version (if known): v12

## Describe the bug
The datastream `vocabularies update` command crashes. Changing a working `vocabularies install` command to `update` doesn't work as expected.

## Steps to Reproduce
1.  Set up a names vocabulary with `pipenv run invenio vocabularies import --vocabulary names --filepath ./vocabularies-future.yaml`

and

`names:
  readers:
    - type: yaml
      args:
        origin: "./app_data/people.yaml"
  writers:
    - type: service
      args:
        service_or_name: names
`
2. Change metadata in the people.yaml file
3. Type `pipenv run invenio vocabularies update --vocabulary names --filepath ./vocabularies-future.yaml`

```
site-packages/invenio_vocabularies/datastreams/writers.py", line 71, in write
    return StreamEntry(self._service.create(self._identity, entry))
  File "site-packages/invenio_records_resources/services/uow.py", line 350, in inner
    res = f(self, *args, **kwargs)
  File "site-packages/invenio_records_resources/services/records/service.py", line 320, in create
    return self._create(self.record_cls, identity, data, uow=uow, expand=expand)
  File "site-packages/invenio_records_resources/services/uow.py", line 354, in inner
    return f(self, *args, **kwargs)
  File "site-packages/invenio_records_resources/services/records/service.py", line 347, in _create
    self.run_components(
  File "site-packages/invenio_records_resources/services/base/service.py", line 72, in run_components
    getattr(component, action)(*args, **kwargs)
  File "site-packages/invenio_vocabularies/services/components.py", line 49, in create
    self.service.record_cls.pid.create(record)
  File "site-packages/invenio_records_resources/records/systemfields/pid.py", line 233, in create
    self.field._provider.create(pid_value, record, self.field.model_field_name)
  File "site-packages/invenio_records_resources/records/providers.py", line 31, in create
    raise PIDAlreadyExists(
invenio_pidstore.errors.PIDAlreadyExists
```

```
site-packages/invenio_vocabularies/cli.py", line 136, in update
    success, errored, filtered = _process_vocab(config)
  File "site-packages/invenio_vocabularies/cli.py", line 69, in _process_vocab
    for result in ds.process():
  File "site-packages/invenio_vocabularies/datastreams/datastreams.py", line 61, in process
    yield self.write(transformed_entry)
  File "site-packages/invenio_vocabularies/datastreams/datastreams.py", line 104, in write
    writer.write(stream_entry)
  File "site-packages/invenio_vocabularies/datastreams/writers.py", line 75, in write
    vocab_id = self._entry_id(entry)
  File "site-packages/invenio_vocabularies/datastreams/writers.py", line 61, in _entry_id
    return (entry["type"], entry["id"])
KeyError: 'type'
```

## Expected behavior

Vocabulary gets updated

Samk13:
- Closes #292
- Fix failing test (unrelated to this change.)

### DIFF
--- a/invenio_vocabularies/cli.py
+++ b/invenio_vocabularies/cli.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020-2021 CERN.
 # Copyright (C) 2021 Graz University of Technology.
+# Copyright (C) 2024 KTH Royal Institute of Technology.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
@@ -131,7 +132,7 @@ def update(vocabulary, filepath=None, origin=None):
     config = get_config_for_ds(vocabulary, filepath, origin)
 
     for w_conf in config["writers"]:
-        w_conf["args"]["update"] = True
+        w_conf.setdefault("args", {})["update"] = True
 
     success, errored, filtered = _process_vocab(config)
 
--- a/tests/resources/test_resources_l10n.py
+++ b/tests/resources/test_resources_l10n.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020-2021 CERN.
 # Copyright (C) 2023 Graz University of Technology.
+# Copyright (C) 2024 KTH Royal Institute of Technology.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
@@ -122,9 +123,22 @@ def test_get(client, example_record, h, prefix, expected_da, expected_en):
 
 def test_search(client, example_record, h, prefix, expected_da, expected_en):
     """Test search result serialization."""
-    expected_en = {"hits": {"hits": [expected_en], "total": 1}}
-    expected_da = {"hits": {"hits": [expected_da], "total": 1}}
 
+    def generate_expected_dict(expected_hits):
+        """Generate expected dict for search results."""
+        return {
+            "hits": {
+                "hits": [expected_hits],
+                "total": 1,
+            },
+            "links": {
+                "self": "https://127.0.0.1:5000/api/vocabularies/resourcetypes2?page=1&size=25&sort=title"
+            },
+            "sortBy": "title",
+        }
+
+    expected_en = generate_expected_dict(expected_en)
+    expected_da = generate_expected_dict(expected_da)
     # Default locale
     res = client.get(f"{prefix}", headers=h)
     assert res.json == expected_en


Samk13:
This change may not address the root issue, which could require further investigation. However, it enables updating vocabulary entries, such as name changes, and ensures updates are applied. 
Note The command will not display the specific changes count in the CLI summary you will get e.g.:
```bash
Vocabulary names updated. Total items 4. 
4 items succeeded <-- should be 1, instead it counts all vocabularies as updated.
0 contained errors
0 were filtered.
```

invenio_vocabularies/cli.py
```diff
@@ -131,7 +132,7 @@ def update(vocabulary, filepath=None, origin=None):
     config = get_config_for_ds(vocabulary, filepath, origin)
 
     for w_conf in config["writers"]:
-        w_conf["args"]["update"] = True
+        w_conf.setdefault("args", {})["update"] = True
```
utnapischtim, line 135: Could this lead to unwanted behavior? 

if the `w_conf["args"]` fails the cli shows the user directly, that the configuration is wrong, but if `setdefault`adds the `args` magically the user didn't know that the `identity`should be set and at best it fails somewhere else or it silently does something the user doesn't want.

Samk13: I agree that using `setdefault` may hide the cause of missing args. What do you think is a better method to handle this you think?  
I can see the [writer_config](https://github.com/inveniosoftware/invenio-vocabularies/blob/a180d29f38f366b37d228314accac96fc74b6d81/invenio_vocabularies/cli.py#L64) coming from [DATASTREAM_CONFIG](https://github.com/inveniosoftware/invenio-vocabularies/blob/a180d29f38f366b37d228314accac96fc74b6d81/invenio_vocabularies/contrib/names/datastreams.py#L118-L125) include [args](https://github.com/inveniosoftware/invenio-vocabularies/blob/a180d29f38f366b37d228314accac96fc74b6d81/invenio_vocabularies/contrib/names/datastreams.py#L121C13-L121C19), should we add `"update": False` to all [DATASTREAM_CONFIG](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-vocabularies%20DATASTREAM_CONFIG%20&type=code) instead?
This change seems to restore functionality, allowing successful vocabulary updates as well. 
But regarding the question of why args is been stipped out while it's there in the config it's not clear for me.

kpsherva: I would rather try to be explicit in the documentation about the arguments rather than setdefault or assume default arguments. I agree that it might lead to unexpected results, and in the cli commands (which are for developer user) we would probably like to avoid unexpected behaviour.

Samk13: Got it! the documentation has been updated accordingly @kpsherva
Then we can close this one once [609](https://github.com/inveniosoftware/docs-invenio-rdm/pull/609) is been merged.

PR #298, opened 2024-02-19
cli : Create factory for get_config_for_ds

Samk13:

### Description

Address the `FIXME` noted in the following section: https://github.com/inveniosoftware/invenio-vocabularies/blob/a180d29f38f366b37d228314accac96fc74b6d81/invenio_vocabularies/cli.py#L27-L51

Preferably merge #297  first to pass the unrelated failing test.

### Checklist

- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

### DIFF
--- a/invenio_vocabularies/factories.py
+++ b/invenio_vocabularies/factories.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+# Copyright (C) 2024 KTH Royal Institute of Technology.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+"""Generate Vocabulary Config."""
+from copy import deepcopy
+
+import yaml
+from invenio_records_resources.proxies import current_service_registry
+
+from .contrib.awards.datastreams import DATASTREAM_CONFIG as awards_ds_config
+from .contrib.funders.datastreams import DATASTREAM_CONFIG as funders_ds_config
+from .contrib.names.datastreams import DATASTREAM_CONFIG as names_ds_config
+
+
+class VocabularyConfig:
+    """Vocabulary Config Factory."""
+
+    config = None
+    vocabulary_name = None
+
+    def get_config(self, filepath=None, origin=None):
+        """Get the configuration for the vocabulary."""
+        config = deepcopy(self.config)
+        if filepath:
+            with open(filepath, encoding="utf-8") as f:
+                config = yaml.safe_load(f).get(self.vocabulary_name)
+        if origin:
+            config["readers"][0].setdefault("args", {})
+            config["readers"][0]["args"]["origin"] = origin
+        return config
+
+    def get_service(self):
+        """Get the service for the vocabulary."""
+        return current_service_registry.get(self.vocabulary_name)
+
+
+class NamesVocabularyConfig(VocabularyConfig):
+    """Names Vocabulary Config."""
+
+    config = names_ds_config
+    vocabulary_name = "names"
+
+
+class FundersVocabularyConfig(VocabularyConfig):
+    """Funders Vocabulary Config."""
+
+    config = funders_ds_config
+    vocabulary_name = "funders"
+
+    def get_service(self):
+        """Get the service for the vocabulary."""
+        raise NotImplementedError("Service not implemented for Funders")
+
+
+class AwardsVocabularyConfig(VocabularyConfig):
+    """Awards Vocabulary Config."""
+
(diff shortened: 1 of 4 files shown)


Samk13:
Question:
in datastrams/dactories L40: https://github.com/inveniosoftware/invenio-vocabularies/blob/a180d29f38f366b37d228314accac96fc74b6d81/invenio_vocabularies/datastreams/factories.py#L28-L42
, specific factories like WriterFactory, ReaderFactory, and TransformerFactory inherit from both `Factory`and `OptionsConfigMixin`, accessing `options()` successfully. 
However, the base Factory class does not inherit from OptionsConfigMixin, leading to Pylint errors when accessing options() `Class 'Factory' has no 'options' memberPylint`. 

Should we adjust the design to include OptionsConfigMixin in Factory for direct options() access, or is the separation intentional to restrict options() access to specific factories only?

invenio_vocabularies/factories.py
```diff
@@ -0,0 +1,97 @@
...
+
+from .contrib.awards.datastreams import DATASTREAM_CONFIG as awards_ds_config
+from .contrib.funders.datastreams import DATASTREAM_CONFIG as funders_ds_config
+from .contrib.names.datastreams import DATASTREAM_CONFIG as names_ds_config
+
+
+class VocabularyConfigFactory:
+    """Base class for Vocabulary Config Factory."""
+
+    def get_config(self, filepath=None, origin=None):
```
utnapischtim, line 23: so third try to create that review:

since the get_config is the same in all three sub classes i would move the code to the base class `VocabularyConfig` and do customization in the sub classes by adding a `vocabulary_name` and `config` property.

```
class VocabularyConfig:
    config = None
    vocabulary_name = None

    def get_config(self, filepath=None, origin=None):
        """Get the configuration for the names vocabulary."""
        # maybe add some checks to help further self to use the subclasses in the right way
        config = deepcopy(self.config)
        if filepath:
            with open(filepath, encoding="utf-8") as f:
                config = yaml.safe_load(f).get(self.vocabulary_name)
        if origin:
            config["readers"][0].setdefault("args", {})
            config["readers"][0]["args"]["origin"] = origin
        return config

    def get_service(self):
        """Get the service for the names vocabulary."""
        return current_service_registry.get(self.vocabulary_name)

class NamesVocabularyConfig:
    config = names_ds_config
    vocabulary_name = "names"
```
i have removed `Factory` from the names, because in my opinion those config classes are not creating anything and are therefore no factory classes

Samk13: @utnapischtim After taking another look, you're absolutely right—thank you for highlighting that. I was probably swayed by the comment on the Fixme...
But if it's not a factory, should we relocate the code back to `cli.py`, `config.py` or is it best to leave it where it currently is?

utnapischtim: btw in the original code only the "names" had a service. Now all vocabs could return a service. Could that be a problem?

Samk13: Good point. We can raise a `NotImplementedError` in the get_service methods for both `funders` and `awards` in that case. Unless you have another approach?

utnapischtim: we could do something like:

```
class VocabularyConfig:
    service_name = None
    
    def get_service(self):
        if self.service_name is None:
            raise NotImplementedError
        return current_service_registry.get(self.service_name)
```
and set the name of the service only in NamesVocabularyConfig

Samk13: I've opted to raise `NotImplementedError` in the subclass get_service methods for  `funders` and `awards`. 
This way, it's unambiguous that these services aren't available yet. Once implemented, we can simply remove the override to enable the service.

jrcastro2 (approved):
LGTM! 
tests are failing, but it migth simply need a rebase.

Issue #307, opened 2024-05-13
Admin panel to list all vocabularies


slint:
![](https://codimd.web.cern.ch/uploads/upload_b20e312a8034c9bd307e5cb0184e1e26.png)

karkraeg:
We implemented a working Admin panel in https://github.com/ulbmuenster/invenio-vocabularies/tree/all_vocabularies_endpoint

![image](https://github.com/inveniosoftware/invenio-vocabularies/assets/8436596/03050e1b-9be5-4e47-a7e7-1227814a6988)

The API is yet to be defined and also we are not sure why one would need the `System` badge?

karkraeg:
missing: 

- `System` Badge
- optional Delete Button

karkraeg:
@slint is it true that we have three types of vocabularies?

- system "custom" vocabularies (names, affiliations, awards, funders) (system badge)
- system "generic" vocabularies (licenses, languages...) (no badge)
- user defined new vocabularies (in the screenshot `code:developmentStatus`) - are deleteable

PR #310, opened 2024-05-14
All vocabularies endpoint and Admin interface

karkraeg:

### Description

Adresses https://github.com/inveniosoftware/invenio-vocabularies/issues/307

### Checklist

- [X] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [X] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [X] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [X] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Frontend**

- [X] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [X] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [X] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**


invenio_vocabularies/resources/config.py
```diff
@@ -0,0 +1,108 @@
...
+    # /vocabulary/vocabulary_id
+    # Blueprint configuration
+    blueprint_name = "vocabulary_list"
+    url_prefix = "/vocabularies"
+
+    routes = {
+        "all": "/",
+        "list": "/vocabularies/<pid_value>",
+        "item": "/vocabularies/<pid_value>/<vocabulary_type_id>",
+    }
```
slint, line 84: minor: won't this produce a `/vocabularies/vocabularies/<pid_value>` endpoint in the end? Though I could see that we're ending up with a name clash with the existing `/vocabularies/<pid_type>` endpoint (e.g. `/vocabularies/licenses`). Maybe we can hardcode having paths like `/vocabularies/types/<type_id>`?

invenio_vocabularies/services/service.py
```diff
@@ -37,6 +47,148 @@
...
+    def search(self, identity):
+        """Search for vocabulary types entries."""
+        self.require_permission(identity, "list_vocabularies")
+
+        vocabulary_types = VocabularyType.query.all()
+
+        # config_vocab_types = current_app.config["INVENIO_VOCABULARY_TYPE_METADATA"]
+        config_vocab_types = current_app.config.get(
+            "INVENIO_VOCABULARY_TYPE_METADATA", {}
+        )
```
slint, line 130: minor/shelve: we could try to do this "response enhancement" in the marshmallow schema or even the service. Since it might change a lot, I would leave it here and postpone though.

slint (approved):
Looks good, some nits and minor comments (and possibly non-important things we shelve in issues).

karkraeg:
depends on https://github.com/inveniosoftware/invenio-administration/pull/208

SarahW91:
I added Uni Münster in the copyright header in the files where we added larger changes. Hope that's okay, otherwise I can remove it again.

invenio_vocabularies/resources/__init__.py
```diff
@@ -9,7 +9,14 @@
 """Resources module."""
 from invenio_vocabularies.resources.schema import L10NString, VocabularyL10Schema
 
+from .config import VocabulariesResourceConfig, VocabularyTypeResourceConfig
+from .resource import VocabulariesAdminResource, VocabulariesResource
```
slint, line 13: Nice, we should try more to expose objects at the top level of the sub-module namespaces.

invenio_vocabularies/resources/resource.py
```diff
@@ -36,56 +40,23 @@
...
-        "application/vnd.inveniordm.v1+json": ResponseHandler(
-            MarshmallowSerializer(
-                format_serializer_cls=JSONSerializer,
-                object_schema_cls=VocabularyL10NItemSchema,
-                list_schema_cls=BaseListSchema,
-            ),
-            headers=etag_headers,
-        ),
-    }
+# from invenio_vocabularies.proxies import current_service
```
slint, line 43:
```suggestion
```

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+    ] + SearchOptions.params_interpreters_cls
+
+    # TODO: Is this still necessary here?
+    suggest_parser_cls = SuggestQueryParser.factory(
+        fields=[
+            "id.text^100",
+            "id.text._2gram",
+            "id.text._3gram",
+        ],
+    )
```
slint, line 105: Probably not necessary anymore since Vocabulary Types are not coming from OpenSearch and follow different filtering rules. I think entirely removing the attributes should be enough and doesn't break anything.

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+        ),
+    }
+
+    sort_default = "id"
+
+    sort_default_no_query = "id"
+
+    sort_direction_options = {
+        "asc": dict(title=_("Ascending"), fn=asc),
+        "desc": dict(title=_("Descending"), fn=desc),
```
slint, line 120:
```suggestion
        "asc": dict(title=_("Ascending"), fn=sa.asc),
        "desc": dict(title=_("Descending"), fn=sa.desc),
```

nit: to keep the namespace cleaner (what I have in mind is `import sqlalchemy as sa`)

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+
+    links_search = pagination_links("{+api}/vocabularies/{type}{?args*}")
+
+
+class VocabularyTypesServiceConfig(RecordServiceConfig):
+    """Vocabulary types service configuration."""
+
+    service_id = "vocabulary_types"
+    permission_policy_cls = PermissionPolicy
+    record_cls = Vocabulary  # TODO: Is this correct?
```
slint, line 165: It should probably be `invenio_vocabularies.records.models:VocabularyType`, though this wouldn't really do change anything (I don't think `.record_cls` is used anywhere for this service).

Long-term we might create a dedicated `VocabularyType` API class under `invenio_vocabularies/records/api.py`, which would also help us with the issues we faced when converting DB objects to dictionaries to fit the service results.

invenio_vocabularies/services/results.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @property
+    def custom_vocabulary_names(self):
+        """Checks whether vocabulary is a custom vocabulary."""
+        return current_app.config.get("VOCABULARIES_CUSTOM_VOCABULARY_TYPES", [])
+
+    def to_dict(self):
+        """Formats result to a dict of hits."""
+        # hits = list(self._results)
+
```
slint, line 37:
```suggestion
```

invenio_vocabularies/services/results.py
```diff
@@ -0,0 +1,112 @@
...
+from flask import current_app
+from invenio_records_resources.proxies import current_service_registry
+from invenio_records_resources.services import RecordServiceConfig
+from invenio_records_resources.services.records.results import RecordList
+from invenio_search import current_search_client
+
+from ..proxies import current_service
+
+
+class VocabularyMetadataList(RecordList):
```
slint, line 21:
```suggestion
class VocabularyTypeList(RecordList):
```

invenio_vocabularies/resources/config.py
```diff
@@ -0,0 +1,109 @@
...
+        ),
+    }
+
+
+class VocabularyTypeResourceConfig(ResourceConfig, ConfiguratorMixin):
+    """Vocabulary list resource config."""
+
+    # /vocabulary/vocabulary_id
+    # Blueprint configuration
+    blueprint_name = "vocabulary_list"
```
slint, line 78:
```suggestion
    blueprint_name = "vocabulary_types"
```

invenio_vocabularies/resources/config.py
```diff
@@ -0,0 +1,109 @@
...
+    url_prefix = "/vocabularies"
+
+    routes = {
+        "all": "/",
```
slint, line 82:
```suggestion
        "all": "",
```

minor: This would also work, and allow just accessing `/api/vocabularies`

invenio_vocabularies/resources/config.py
```diff
@@ -0,0 +1,109 @@
...
+        "list": "/vocabularies/<pid_value>",
+        "item": "/vocabularies/<pid_value>/<vocabulary_type_id>",
```
slint, line 84: minor: I'm not sure if these two endpoints are used anywhere right now. I.e. if I request `/api/vocabularies/vocabularies/subjects` I'm not sure what response would come back.

Worst case we can just remove them, in which case we could also replace `"all"` with `"list"` in this context.

invenio_vocabularies/resources/resource.py
```diff
@@ -164,3 +135,37 @@ def launch(self):
...
+    """Resource for vocabularies admin interface."""
+
+    def create_url_rules(self):
+        """Create the URL rules for the record resource."""
+        routes = self.config.routes
+        rules = super().create_url_rules()
+
+        rules.append(
+            route("GET", routes["all"], self.get_all_vocabulary_types),
+        )
```
slint, line 150: minor: if we replace `"all"` with `"list"`, then we can completely remove this method actually, and rename `get_all_vocabulary_types` to `search` (so that it maps correctly to the `GET /api/vocabularies` route rule.

invenio_vocabularies/resources/resource.py
```diff
@@ -164,3 +135,37 @@ def launch(self):
...
+
+    @request_view_args
+    @response_handler()
+    def read(self):
+        """Read an item."""
+        pid_value = (
+            resource_requestctx.view_args["type"],
+            resource_requestctx.view_args["pid_value"],
+        )
+        item = self.service.read(g.identity, pid_value)
```
slint, line 170: minor: not sure if this will work without overriding/defining `VocabularyTypeService.read(...)` to use the parameters to fetch the vocabulary type from the DB.

On the other hand, I'm not sure if this is also used anywhere.

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+    schema = VocabularySchema
```
slint, line 166:
```suggestion
    schema = VocabularyTypeSchema
```

minor: I don't think we have it defined though, in which case we can also skip it entirely

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+    task_schema = TaskSchema
```
slint, line 167:
```suggestion
```

This is only needed in `VocabulariesServiceConfig`.

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+    vocabularies_listing_resultlist_cls = results.VocabularyMetadataList
+
+    vocabularies_listing_item = {
```
slint, line 170:
```suggestion
    result_list_cls = results.VocabularyMetadataList

    links_item = {
```

minor: since it's the only type of result we have in the service

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,205 @@
...
+        "self": Link(
+            "{+api}/vocabularies/{type}/{id}",
+            vars=lambda record, vars: vars.update(
+                {
+                    "id": record.pid.pid_value,
+                    "type": record.type.id,
+                }
+            ),
+        ),
+    }
```
slint, line 203: minor: I'm not sure this is used anywhere by the service.

invenio_vocabularies/services/service.py
```diff
@@ -20,97 +20,64 @@
...
+    def schema(self):
+        """Returns the data schema instance."""
+        return ServiceSchemaWrapper(self, schema=self.config.schema)
+
+    @property
+    def links_item_tpl(self):
+        """Item links template."""
+        return LinksTemplate(
+            self.config.vocabularies_listing_item,
+        )
```
slint, line 45:
```suggestion
```

With some of the changes suggested in the config, these will not be needed anymore.

slint (approved):
Thanks a lot for the clean-up and splitting up the resource and service configs. I think this is looking good and most of my comments are about naming or removing duplicated or redundant config vars and attributes.

I know it's a lot of comments, so if you prefer we could also organize a quick call and go through some of the heavier ones together.

invenio_vocabularies/resources/config.py
SarahW91: Maybe we can look at this together. I am not quite comfortable removing these without knowing all the possible implications

invenio_vocabularies/resources/resource.py
SarahW91: see above comment

invenio_vocabularies/resources/resource.py
SarahW91: How could I check if this works?

invenio_vocabularies/services/config.py
SarahW91: We don't have it defined currently. Does "skip" mean I can simply remove this line?

invenio_vocabularies/services/config.py
SarahW91: There is actually another links_item below - how should I treat this?

invenio_vocabularies/resources/config.py
SarahW91: This does lead to 404 errors when the endpoint is called from the frontend, because it uses e.g. `/api/vocabularies/?q=&page=1&size=10` with a slash

SarahW91: Calling the URL without a slash just redirects currently, so both versions of URL work with the current code

invenio_vocabularies/services/service.py
SarahW91: The view does not render without the schema property present.

invenio_vocabularies/resources/resource.py
SarahW91: The parent's create_url_rules uses routes["items"] though, so we still need to overwrite this method in the VocabulariesAdminResource which uses "list".

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class VocabularySearchOptions(SearchOptions):
+    """Search options for vocabularies."""
+
+    params_interpreters_cls = [
+        FilterParam.factory(param="tags", field="tags"),
+    ] + SearchOptions.params_interpreters_cls
+
+    suggest_parser_cls = SuggestQueryParser.factory(
```
Samk13, line 56: Question:
Can you explain why these specific fields and weights boosting were chosen for the suggest_parser_cls?
Documenting the rationale behind this decision for future reference would be helpful as a small comment in the class docstring.
see comments here: https://github.com/inveniosoftware/invenio-vocabularies/pull/347#discussion_r1662471284

SarahW91: Since I just moved the code in this PR and did not write it I'll leave answering this question to someone with more expertise on this ;)

PR #329, opened 2024-05-27
first implementation of OAIPMHReader

wgresshoff:

### Description

Add a new reader for harvesting of OAI-PMH based vocabularies like those at GND.

### Checklist

- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**


invenio_vocabularies/datastreams/models.py
slint, line 1: minor/nit: I would "inline" this file in the OAI-PMH reader, since it's tightly bound and probably the only place where it's going to be used.

I know the file is growing somewhat large, and the way to properly clean up would be to split more complex reader types into separate files while keeping all their logic/API classes together.

setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     invenio-records-resources>=5.0.0,<6.0.0
     lxml>=4.5.0
     PyYAML>=5.4.1
+    invenio-oaipmh-scythe @ git+https://github.com/ulbmuenster/invenio-oaipmh-scythe.git
```
slint, line 34: minor/shelve: we can move the fork under `inveniosoftware` as well once we have a fully working datastream example :)

invenio_vocabularies/datastreams/readers.py
```diff
@@ -222,3 +225,56 @@ def _iter(self, fp, *args, **kwargs):
...
+                from_=self._from,
+                until=self._until,
+                metadata_prefix=self._metadata_prefix,
+                set_=self._set,
+                ignore_deleted=True,
+            )
+            for response in responses:
+                oaipmh_record = {
+                    "header": response.header,
+                    "metadata": response.metadata["record"],
```
slint, line 267: minor/question: is `response.header` an XML tree Python object, or a custom internal class of Scythe?

It might be an option to also just return `{"response": response}`, and leave it to the transformer (or the next reader) to extract whatever information is needed.

slint (approved):
Looks good and very much in a working state. Some minor comments on organization and how to handle the new dependency.

invenio_vocabularies/datastreams/models.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# Copyright (C) 2024 University of Münster.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Record Model for Scythe."""
+
+from invenio_oaipmh_scythe.models import Record
```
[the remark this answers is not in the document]
wgresshoff: Is it possible to use the same package name? I changed the name to have a different package in PyPi.

tests/datastreams/test_datastreams.py
```diff
@@ -144,3 +145,60 @@ def test_piping_readers(app, zip_file, json_element):
...
     assert count == 5  # 2 good + 1 bad + 2 good
+
+
+def test_oaipmh_reader(app):
+    ds_config = {
+        "readers": [
+            {
+                "type": "oai-pmh",
+                "args": {
+                    "base_url": "https://services.dnb.de/oai/repository",
```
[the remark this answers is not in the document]
wgresshoff: I was aware of the danger but wanted a quick solution. I have to investigate the possibilities. Possibly using pytest-httpserver may help here.

setup.cfg
wgresshoff: Sure, if we don't need a Pypi package we should.

invenio_vocabularies/datastreams/readers.py
wgresshoff: Yes, `response.header` is an internal class (looked simply like a string at first look), as is `response`, too. If that's no problem you're right. I'll change that.

invenio_vocabularies/datastreams/models.py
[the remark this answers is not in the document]
wgresshoff: This is changed.

invenio_vocabularies/datastreams/models.py
wgresshoff: Done. Regarding the split into several files: I think so, too. We handle refactorings like this in different issues and PR's, what do you think?

tests/datastreams/test_datastreams.py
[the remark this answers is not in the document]
wgresshoff: I replaced the code with a mocking solution based on `pytest_httpserver`.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -222,3 +226,80 @@ def _iter(self, fp, *args, **kwargs):
...
+            for record in records:
+                yield {"record": record}
+        except NoRecordsMatch:
+            raise ReaderError(f"No records found in OAI-PMH request.")
+
+    def read(self, item=None, *args, **kwargs):
+        """Reads from item or opens the file descriptor from origin."""
+        if item:
+            pass
+            # this will never be called!
```
ptamarit, line 274:
```suggestion
            raise NotImplementedError(
                "OAIPMHReader does not support being chained after another reader"
            )
```

This is what I did in the two readers I recently implemented, so that if somebody mistakenly defines a reader before this one, they will get a clear exception instead of silently doing nothing.

setup.cfg
```diff
@@ -31,12 +31,14 @@ install_requires =
     invenio-records-resources>=5.0.0,<6.0.0
     lxml>=4.5.0
     PyYAML>=5.4.1
+    oaipmh-scythe @ git+https://github.com/ulbmuenster/invenio-oaipmh-scythe.git
```
ptamarit, line 34: Do we want to release `invenio-oaipmh-scythe` to PyPI before merging or are we happy to keep a Git dependency?

setup.cfg
```diff
@@ -31,12 +31,14 @@ install_requires =
...
 
 [options.extras_require]
 tests =
     pytest-black-ng>=0.4.0
     invenio-app>=1.4.0,<2.0.0
     invenio-db[postgresql,mysql]>=1.0.14,<2.0.0
+    pytest_httpserver>=1.0.10
```
ptamarit, line 41: @slint had mentioned a few libraries in the [`RORHTTPReader` Pull Request](https://github.com/inveniosoftware/invenio-vocabularies/pull/315#discussion_r1608132835).
I guess the decision on which library to use does not matter too much (since it's test-only), but I was curious about how you chose this specific one.

tests/datastreams/test_datastreams.py
```diff
@@ -144,3 +146,154 @@ def test_piping_readers(app, zip_file, json_element):
...
+                    "base_url": httpserver.url_for("/oai/repository"),
+                    "metadata_prefix": "MARC21plus-1-xml",
+                    "set": "authorities:sachbegriff",
+                    "from_date": "2024-01-01T09:00:00Z",
+                    "until_date": "2024-01-31T10:00:00Z",
+                },
+            },
+        ],
+        "writers": [{"type": "test"}],
+    }
```
ptamarit, line 251: Since you are testing `OAIPMHReader` specifically, I would say that defining a full config/factory here is not necessary.
You could:
1. Move these tests from `tests/datastreams/test_datastreams.py` to `tests/datastreams/test_readers.py`
2. Change the test to look more like other reader tests (calling directly the constructor and then the `read` method).
3. Remove the definition of the OAIPMHReader in `tests/datastreams/conftest.py`

tests/datastreams/test_datastreams.py
```diff
@@ -144,3 +146,154 @@ def test_piping_readers(app, zip_file, json_element):
...
+
+    datastream = DataStreamFactory.create(
+        readers_config=ds_config["readers"],
+        writers_config=ds_config["writers"],
+    )
+    iter = datastream.process()
+    for count, entry in enumerate(iter, start=1):
+        assert "record" in entry.entry
+        break
```
ptamarit, line 260: I'm not understanding why we use a `for` loop is it's to directly `break` out of it.

tests/datastreams/test_datastreams.py
```diff
@@ -144,3 +146,154 @@ def test_piping_readers(app, zip_file, json_element):
...
+    datastream = DataStreamFactory.create(
+        readers_config=ds_config["readers"],
+        writers_config=ds_config["writers"],
+    )
+    stream_iter = datastream.process()
+    try:
+        _ = next(stream_iter)
+        assert False
+    except ReaderError:
+        assert True
```
ptamarit, line 299:
```suggestion
    with pytest.raises(ReaderError):
        next(stream_iter)
```

This is what I used in a few tests called `test_ror_http_reader_...`, I think it should work.

ptamarit (approved):
Looks good to me.
I have a few minor comments, and I'm also _wondering if all the commits should be squashed into one_.

setup.cfg
wgresshoff: I thought so, too, but it's just a temporary solution until we completely get rid of Python 3.9.

setup.cfg
wgresshoff: I did not look too deep for a library. `Response` (in my opinion) is heavily written for testings based on `requests`, while `oaipmh_scythe` is based on `httpx`. It seemed a good idea to use a library that directly hooks into the protocol, thus working independently from any library used to implement http calls.

tests/datastreams/test_datastreams.py
wgresshoff: Done.

PR #330, opened 2024-05-28
global: add "tags" field to all vocabularies

slint:


tmorrell (approved):
Looks good, and should rebase fine. In ROR I put the analogous info in the "status" field. I'm thinking it's fine to have both a vocabulary-specific status as well as a standard RDM "tag". That way RDM can apply rules across all vocabularies with tags, and we could do ROR-specific behavior as well if we need to.

ptamarit:
Similar to funders on ROR v2 adding mapping fields but not breaking functionalities.

PR #332, opened 2024-05-28
Subjects to datastreams

karkraeg:

### Description

I started with what I was thinking were essential in porting subjects to datastreams. There already has been a `YamlReader` and a `CSVReader` which we will use to keep backwards compatibility to the current way of loading subjects from files. I added a `SubjectsYAMLTransformer` that is somehow redundant because the dict returned by the YAML reader is basically the same as what the reader returns. But it might me useful at some point I guess?

I thinks whats missing is the implementation of a writer method similar to `VocabularyEntryWithSchemes` in invenio-rdm-records (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/vocabularies.py#L383)

### DIFF
--- a/invenio_vocabularies/contrib/subjects/datastreams.py
+++ b/invenio_vocabularies/contrib/subjects/datastreams.py
@@ -45,23 +45,6 @@ class YAMLTransformer(BaseTransformer):
         for synonym in record["synonyms"]:
             subject["synonyms"].append(synonym)
 
-        if self.vocab_schemes:
-            valid_schemes = set(self.vocab_schemes.keys())
-        else:
-            valid_schemes = set()
-        subject["identifiers"] = []
-        for identifier in record["identifiers"]:
-            schemes = idutils.detect_identifier_schemes(identifier)
-            for scheme in schemes:
-                if scheme in valid_schemes:
-                    subject["identifiers"].append(
-                        {
-                            "identifier": identifier,
-                            "scheme": scheme,
-                        }
-                    )
-                    break
-
         return subject
 
 
--- a/invenio_vocabularies/contrib/subjects/datastreams.py
+++ b/invenio_vocabularies/contrib/subjects/datastreams.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021-2022 CERN.
+#
+# Invenio-Vocabularies is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Names datastreams, transformers, writers and readers."""
+
+from invenio_access.permissions import system_identity
+from invenio_records.dictutils import dict_lookup
+
+from ...datastreams.errors import TransformerError
+from ...datastreams.readers import SimpleHTTPReader, BaseReader
+from ...datastreams.transformers import BaseTransformer
+from ...datastreams.writers import ServiceWriter
+
+
+class SubjectsYAMLReader(BaseReader):
+    """Subjects YAML Reader."""
+
+    def __init__(self, *args, options=None, **kwargs):
+        """Constructor."""
+        self._options = options or {}
+        super().__init__(*args, **kwargs)
+
+    def _iter(self, data, *args, **kwargs):
+        """Iterates over a dictionary and returns a dictionary per element."""
+        for entry in data:
+            yield entry
+
+    def read(self, item=None, *args, **kwargs):
+        """Read the YAML file."""
+        import yaml
+
+        with open(self._options["filename"], "r") as f:
+            data = yaml.safe_load(f)
+            yield from self._iter(data=data, *args, **kwargs)
+
+
+class OAIPMHSubjectsReader(BaseReader):
+    # not implemented
+    pass
+
+
+class SubjectsYAMLTransformer(BaseTransformer):
+    """Subjects Transformer."""
+
+    def apply(self, stream_entry, **kwargs):
+        """Transform the data to the invenio subjects format."""
+        record = stream_entry.entry
+        subject_entry = {}
+        print("Transforming data to invenio subjects format...")
+
+        subject_entry["subject"] = record.get("subject")
+        subject_entry["identifier"] = record.get("identifier")
+        subject_entry["scheme"] = record.get("scheme")
+        subject_entry["parent"] = record.get("parent")
+
+        return subject_entry
+
+
+class SubjectsServiceWriter(ServiceWriter):
+    """Subjects Service Writer."""
+
+    def __init__(self, *args, **kwargs):
+        """Constructor."""
+        super().__init__(*args, **kwargs)
+
+    def write(self, data, *args, **kwargs):
+        """Write the data to the service."""
+
+
+VOCABULARIES_DATASTREAM_TRANSFORMERS = {
+    "yaml-vocabulary": SubjectsYAMLTransformer,
+}
+"""ROR Data Streams transformers."""
+
+
+VOCABULARIES_DATASTREAM_WRITERS = {
+    "subjects-service": SubjectsServiceWriter,
+}
+"""Funders Data Streams transformers."""
+
+
+DATASTREAM_CONFIG = {
+    "readers": [
+        {"type": "yaml"},
+    ],
+    "transformers": [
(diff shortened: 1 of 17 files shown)


invenio_vocabularies/services/service.py
```diff
@@ -37,6 +37,147 @@
...
 
 
+def is_custom_vocabulary_type(vocabulary_type, context):
+    """Check if the vocabulary type is a custom vocabulary type."""
+    return vocabulary_type["id"] in current_app.config.get(
+        "VOCABULARIES_CUSTOM_VOCABULARY_TYPES", []
+    )
+
+
+# class VocabularyMetadataList(ServiceListResult):
```
fenekku, line 47: Commented out. Still necessary?

invenio_vocabularies/contrib/subjects/schema.py
```diff
@@ -23,13 +33,33 @@ class SubjectSchema(BaseVocabularySchema):
...
+    synonyms = fields.List(SanitizedUnicode())
+    identifiers = IdentifierSet(
+        fields.Nested(
+            partial(
+                IdentifierSchema,
+                allowed_schemes=subject_schemes,
+                identifier_required=False,
+            )
+        )
+    )
```
fenekku, line 46: Hmmm... I think I lack the context for this change. 

My initial reaction would be:

- yes for the idea of capturing the translated labels
- no for almost anything else because too much change at once :upside_down_face: 

Detail
- Why synonyms? With 1M+ subjects for us, it's already near impossible to surface the right one with current UI interface. If different UI interface perhaps... but if loosey (non-controlled vocab), then let users use keywords. The amount of data and sourcing for this makes me weary too...
- identifiers: are those equivalent identifiers in other vocabularies? What source for these? If one has LCSH and GND subjects, do both hold that information? this duplicates data tremendously (just with 2 subjects we would get at least 2x usage that grows exponentially from there...) + how to keep in sync. If no duplication, which is the main subject?

My take:

If we do want this additional data, I think we need to extract the concept represented by these additions into its own thing: a new resource called "SubjectEquivalencies" or "SubjectSets" where each entry captures that locus of equivalent subjects/keywords. No duplication, clear access, clear concept. But again this might be much for the original task to just move to datastreams.

tests/contrib/subjects/conftest.py
```diff
@@ -21,9 +22,33 @@
...
     return {
-        "id": "https://id.nlm.nih.gov/mesh/D000001",
-        "scheme": "MeSH",
-        "subject": "Calcimycin",
+        "subject": {
+            "en": "Dark Web",
+            "de": "Darknet",
+            "fr": "Réseaux anonymes (informatique)",
+        },
+        "id": "1062531973",
```
fenekku, line 30: `id` has to be a URI. Subjects are not "scheme-spaced" they are all in the same container so they have to be distinct from any other future potential subject id.

fenekku (commented):
Unfortunately I don't have much context for these changes and I feel like I am just a party pooper by commenting. But might as well share my thoughts since I was asked. 

My big takeaways would be:
- a bit too much at once here
- focus on just the datastream in 1 PR (no change to the metadata model)
- then in a separate PR focus on metadata change
- Strongly consider separation of concepts: Subjects from SubjectsEquivalents

invenio_vocabularies/contrib/subjects/datastreams.py
```diff
@@ -0,0 +1,114 @@
...
+from invenio_access.permissions import system_identity
+from invenio_i18n import lazy_gettext as _
+
+from ...datastreams.errors import TransformerError
+from ...datastreams.transformers import BaseTransformer
+from ...datastreams.writers import ServiceWriter
+from .config import subject_schemes
+
+
+class YAMLTransformer(BaseTransformer):
```
fenekku, line 22: Also: YAML is ... actually kind of terrible as an data interchange format. Switching to CSV for my the subjects I manage has cut load time by 99% (literally not kidding) and space by around 66%. I perfectly understand this is what datastreams deal with originally, so it's not "a choice" in this work initially and I don't expect adjustments for that. Just sharing is caring. (this is also in part why I never bothered with datastreams...)

invenio_vocabularies/contrib/subjects/mappings/os-v1/subjects/subject-v1.0.0.json
```diff
@@ -56,6 +59,19 @@
...
+      "identifiers": {
+        "properties": {
+          "identifier": {
+            "type": "keyword"
+          },
+          "scheme": {
+            "type": "keyword"
+          }
+        }
+      },
```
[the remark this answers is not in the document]
wgresshoff: There are two use cases:
- in many places (LCSH, GND, ...) the identifier of a subject is not URI-based so we may have here the identifier as non URI version and can add the URI based one (for metadata exports?!) in the identifiers field.
- the GND has often linked in equivalences from other subject headings like LCSH or Rameau. May help national libraries when harvesting our data.

invenio_vocabularies/contrib/subjects/datastreams.py
wgresshoff: Yes, I see. We will replace this locally with a Marc21Transformer. It's just some sort of "base implementation", a change is left to another PR.

tests/contrib/subjects/conftest.py
wgresshoff: I understand your point and wasn't aware of this. So far we had some subjects (like DDC) where no URI based id was available. This has to be clarified in the docs because this is very important for the implementation of export formats (e.g. for use with OAI-PMH -> most format largely depends on id's).

invenio_vocabularies/contrib/subjects/schema.py
wgresshoff: Synonyms are useful in case of abbreviations which are not of subject headings on their own (e.g. "European Union" vs. "EU"). As you never know which exact term is part of the subject headings you have a better chance to find it by synonyms (if available) or translations. So the key is: it's not more subjects but just better chances to find the correct one in case you don't know the exact term.
Regarding the identifiers, perhaps it's just overdone.

invenio_vocabularies/contrib/subjects/mappings/os-v1/subjects/subject-v1.0.0.json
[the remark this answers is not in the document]
wgresshoff: After some internal discussion we concluded this as useless. I have removed the identifiers agin.

invenio_vocabularies/contrib/subjects/schema.py
```diff
@@ -23,13 +33,33 @@ class SubjectSchema(BaseVocabularySchema):
     # here.
     id = SanitizedUnicode(required=True)
     scheme = SanitizedUnicode(required=True)
-    subject = SanitizedUnicode(required=True)
+    subject = i18n_strings
```
tmorrell, line 36: Do we want to switch this to `title` to be consistent with other vocabularies, or add the il8n_strings version as "title"? I think going from a string to il8n_strings is already a breaking change, so maybe we take the opportunity to get more consistency across vocabularies.

slint: Actually, [`title` already exists in `BaseVocabularySchema`](https://github.com/inveniosoftware/invenio-vocabularies/blob/810583f1354b997ce57101030a7631116e031802/invenio_vocabularies/services/schema.py#L90) as an `i18n_strings` field. In that case, @wgresshoff would it make sense that we leave `subject` as is, and instead populate translations in the `title` field? 

This also achieves the following:

- makes this change backward compatible (so no need to bump the subject schema/mappings to v2.0.0)
  - I only see new fields: `synonyms`, `identifiers` (is this still valid after [this comment](https://github.com/inveniosoftware/invenio-vocabularies/pull/332#discussion_r1647580257)?), and `title`
  - `title` needs to be added still to the JSONSchema and mappings thought
- we align with what we discussed with @SarahW91 today about making it possible for the administration UI to have a stable field to use as a localized display string

wgresshoff:
Has been replaced by https://github.com/inveniosoftware/invenio-vocabularies/pull/355 so I close it.

PR #344, opened 2024-06-05
tasks: fixed exception logging

alejandromumo:


ptamarit (approved):
We talked about potentially counting exceptions and only logging one at the end, but let's see first how this change behaves.

Issue #346, opened 2024-06-26
Admin panel: Add Vocabulary item list view


SarahW91:
Implement vocabulary item list view as described in https://codimd.web.cern.ch/UTnvE-L2RrSar3_szWWK_w

![](https://codimd.web.cern.ch/uploads/upload_c8fa3194c384e1e1c78450e2ae200bd9.png)

SarahW91:
I still have a few issues with this view:
- Multiple columns are shown for all possible name/title variations. It would be nicer to select the appropriate one(s) depending on the vocabulary type.
- Title is always used in English instead of preferring the instances selected language.
- Search only works via ID of an item, not the displayed name/title.
- Sorting by anything other than the displayed name/title leads to an error (and sort doesn't work in the UI anyways, does it?)

SarahW91:
@slint Most of my questions were resolved with the changes we discussed last week. However, the item search is still a bit weird to me. Searching for partial names/titles/subjects sometimes leads to results, sometimes not. I don't quite understand why. 
Another thing is the localization of the displayed title. I'm not quite sure if this can be done or how.

PR #347, opened 2024-06-26
Vocabulary item view

SarahW91:
### Description

Addresses #346 

### Checklist

**Frontend**

- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_vocabularies/administration/views/vocabularies.py
+++ b/invenio_vocabularies/administration/views/vocabularies.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2024 CERN.
+# Copyright (C) 2020-2021 CERN.
 # Copyright (C) 2024 Uni Münster.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
@@ -8,13 +8,11 @@
 # details.
 
 """Vocabularies admin interface."""
-from flask import current_app
 from functools import partial
-from invenio_search_ui.searchconfig import search_app_config, SortConfig, FacetsConfig
-from invenio_administration.views.base import (
-    AdminResourceEditView,
-    AdminResourceListView,
-)
+
+from flask import current_app
+from invenio_administration.views.base import AdminResourceListView
+from invenio_search_ui.searchconfig import FacetsConfig, SortConfig, search_app_config
 
 
 class VocabulariesListView(AdminResourceListView):
@@ -62,7 +60,9 @@ class VocabularyDetailsListView(AdminResourceListView):
     def init_search_config(self, **kwargs):
         """Build search view config."""
         pid_value = kwargs.get("pid_value", "")
-        custom_search_config = current_app.config[self.search_config_name].get(pid_value)
+        custom_search_config = current_app.config[self.search_config_name].get(
+            pid_value
+        )
 
         if custom_search_config:
             available_sort_options = current_app.config[self.search_sort_config_name]
@@ -75,10 +75,12 @@ class VocabularyDetailsListView(AdminResourceListView):
                 sort_options=available_sort_options,
                 endpoint=self.get_api_endpoint(**kwargs),
                 headers=self.get_search_request_headers(**kwargs),
-                sort=SortConfig(available_sort_options,
-                                custom_search_config["sort"],
-                                custom_search_config["sort_default"],
-                                custom_search_config["sort_default_no_query"]),
+                sort=SortConfig(
+                    available_sort_options,
+                    custom_search_config["sort"],
+                    custom_search_config["sort_default"],
+                    custom_search_config["sort_default_no_query"],
+                ),
                 facets=FacetsConfig(available_facets, custom_search_config["facets"]),
             )
         else:
@@ -96,9 +98,9 @@ class VocabularyDetailsListView(AdminResourceListView):
         if custom_config:
             parent_context.update(custom_config)
         else:
-            parent_context.update({
-                "title": f"{pid_value.capitalize()} vocabulary items"
-            })
+            parent_context.update(
+                {"title": f"{pid_value.capitalize()} vocabulary items"}
+            )
 
         return self.render(**parent_context)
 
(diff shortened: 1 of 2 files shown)


SarahW91:
This is based on #310 and that PR should probably be merged first.
I have some questions left that I added as TODO comments in the code - whoever reviews this: would be great if you could help me out with those.

invenio_vocabularies/services/config.py
```diff
@@ -0,0 +1,181 @@
...
+
+class VocabularySearchOptions(SearchOptions):
+    """Search options for vocabularies."""
+
+    params_interpreters_cls = [
+        FilterParam.factory(param="tags", field="tags"),
+    ] + SearchOptions.params_interpreters_cls
+
+    suggest_parser_cls = SuggestQueryParser.factory(
+        fields=[
```
Samk13, line 57: Question:
Can you explain why these specific fields and weights boosting were chosen for the suggest_parser_cls? 
Documenting the rationale behind this decision for future reference would be helpful as a small comment in the class docstring.

SarahW91: I did not add this code, so someone else would have to answer the question. I simply split the service and the Config into two separate files (this is actually done in #310, that's why I recommend merging that PR first and then rebasing this branch).

invenio_vocabularies/administration/views/vocabularies.py
```diff
@@ -0,0 +1,118 @@
...
+        "id": {"text": "Name", "order": 1},
+        "count": {"text": "Number of entries", "order": 2},
+    }
+
+    search_config_name = "VOCABULARIES_TYPES_SEARCH"
+    search_facets_config_name = "VOCABULARIES_TYPES_FACETS"
+    search_sort_config_name = "VOCABULARIES_TYPES_SORT_OPTIONS"
+
+
+class VocabularyDetailsListView(AdminResourceListView):
```
kpsherva, line 45: I noticed DetailView inherits from ListView here, is there a reason? I think it might cause some issues

invenio_vocabularies/administration/views/vocabularies.py
```diff
@@ -0,0 +1,118 @@
...
+    def get_api_endpoint(self, pid_value=None):
+        """Overwrite get_api_endpoint to accept pid_value."""
+        if pid_value in current_app.config.get(
+            "VOCABULARIES_CUSTOM_VOCABULARY_TYPES", []
+        ):
+            return f"/api/{pid_value}"
+        else:
+            return f"/api/vocabularies/{pid_value}"
+
+    def get_context(self, **kwargs):
```
kpsherva, line 57: the parent class method which generates the context is defined as `.get`. If understand correctly, this method is meant to override it, shouldn't it have the same name? Unless I am missing some history here

invenio_vocabularies/administration/views/vocabularies.py
```diff
@@ -0,0 +1,118 @@
...
+            "display_read": self.display_read,
+            "actions": self.serialize_actions(),
+            "pid_path": self.pid_path,
+            "pid_value": pid_value,
+            "create_ui_endpoint": self.get_create_view_endpoint(),
+            "list_ui_endpoint": self.get_list_view_endpoint(),
+            "resource_name": (
+                self.resource_name if self.resource_name else self.pid_path
+            ),
+        }
```
kpsherva, line 82: please check my previous comment about function name, it is related to the suggestion below
We could avoid some of the duplication below if we use parent class method

```suggestion
        parent_context = super().get(**kwargs)

        pid_value = kwargs.get("pid_value", "")
        return parent_context.update(
            "title": f"{pid_value.capitalize()} vocabulary items",
            "pid_value": pid_value,
        }
```

invenio_vocabularies/administration/views/vocabularies.py
SarahW91: Not yet. Could you explain how that might help with our issues?

invenio_vocabularies/services/config.py
SarahW91: I'll resolve this conversation for this PR, since it is not related to changes made here. If there is a problem with indexing you could create a new issue.

invenio_vocabularies/administration/views/vocabularies.py
kpsherva: it would ensure less ambiguity on where the methods are actually coming from, this class now only imitates the interface of AdminResourceDetails view in order to use some of its frontend functionalities, while it does not actually inherit from the class itself. It might create divergence on the implementation of these "virtually" inherited methods the code it will be harder to maintain.
So to answer your question, it does not solve immediate problems but prevents ambiguity in the future

invenio_vocabularies/administration/views/vocabularies.py
```diff
@@ -54,45 +57,64 @@ def get_api_endpoint(self, pid_value=None):
...
-            "actions": self.serialize_actions(),
-            "pid_path": self.pid_path,
-            "pid_value": pid_value,
-            "create_ui_endpoint": self.get_create_view_endpoint(),
-            "list_ui_endpoint": self.get_list_view_endpoint(),
-            "resource_name": (
-                self.resource_name if self.resource_name else self.pid_path
-            ),
-        }
+        vocab_admin_cfg = current_app.config["VOCABULARIES_ADMINISTRATION_CONFIG"]
```
kpsherva, line 94: I would like to suggest using class factory pattern here instead of moving it to a static dict config in the configuration variable. I understand the dict is imitating the AdminView class interface but if it is not really a class, then we diverge of the pattern of implementation of the administration module

invenio_vocabularies/administration/views/vocabularies.py
SarahW91: Sure, no problem

invenio_vocabularies/administration/views/vocabularies.py
SarahW91: I talked about this with @slint and he suggested doing it this way. I understand your concerns but I'm not sure how I could implement this the way you suggest.

PR #362, opened 2024-07-15
installation: use invenio-oaipmh-scythe from PyPI

slint:


yashlamba (approved):
CI's approval is my approval 👍 :shipit:

Issue #353, opened 2024-07-08
ORCID Public Data Sync differential import


jrcastro2:
* ([See CERN's premium credentials details](https://github.com/inveniosoftware/invenio-vocabularies/issues/314))
  * ([Official example Python code](https://github.com/ORCID/public-data-sync))
  * Define a reader which finds entries to be updated based on last successful run or which imports everything
  * Reuse or define the transformer and the writer
  * Define the celery task in the job system and run it
      * Check performance, see if the long-running job successfully finishes
      * In this case, upading existing records is fine since we do not need to propagate updates (records keep the names entered at the moment of the publication)

PR #366, opened 2024-07-19
Add orcid data sync

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-vocabularies/issues/353

### DIFF
--- a/invenio_vocabularies/config.py
+++ b/invenio_vocabularies/config.py
@@ -155,3 +155,14 @@ VOCABULARIES_TYPES_SEARCH = {
     "sort": ["name", "count"],
 }
 """Vocabulary type search configuration."""
+
+VOCABULARIES_ORCID_ACCESS_KEY = "TODO"
+"""ORCID access key to access the s3 bucket."""
+VOCABULARIES_ORCID_SECRET_KEY = "TODO"
+"""ORCID secret key to access the s3 bucket."""
+VOCABULARIES_ORCID_SUMMARIES_BUCKET = "v3.0-summaries"
+"""ORCID summaries bucket name."""
+VOCABULARIES_ORCID_SYNC_MAX_WORKERS = 32
+"""ORCID max number of simultaneous workers/connections."""
+VOCABULARIES_ORCID_SYNC_DAYS = 1
+"""ORCID number of days to sync."""
--- a/invenio_vocabularies/config.py
+++ b/invenio_vocabularies/config.py
@@ -164,5 +164,7 @@ VOCABULARIES_ORCID_SUMMARIES_BUCKET = "v3.0-summaries"
 """ORCID summaries bucket name."""
 VOCABULARIES_ORCID_SYNC_MAX_WORKERS = 32
 """ORCID max number of simultaneous workers/connections."""
-VOCABULARIES_ORCID_SYNC_DAYS = 1
-"""ORCID number of days to sync."""
+VOCABULARIES_ORCID_SYNC_SINCE = {
+    "days": 1,
+}
+"""ORCID time shift to sync. Parameters accepted are the ones passed to 'datetime.timedelta'."""
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -12,10 +12,11 @@ import csv
 import io
 import tarfile
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from datetime import datetime, timedelta
+from datetime import timedelta
 
 import arrow
 from flask import current_app
+from invenio_access.permissions import system_identity
 from invenio_records.dictutils import dict_lookup
 
 from invenio_vocabularies.contrib.names.s3client import S3OrcidClient
@@ -29,7 +30,7 @@ from ...datastreams.writers import ServiceWriter
 class OrcidDataSyncReader(BaseReader):
     """ORCiD Data Sync Reader."""
 
-    def __init__(self, origin=None, mode="r", *args, **kwargs):
+    def __init__(self, origin=None, mode="r", since=None, *args, **kwargs):
         """Constructor.
 
         :param origin: Data source (e.g. filepath).
@@ -37,6 +38,7 @@ class OrcidDataSyncReader(BaseReader):
         """
         super().__init__(origin=origin, mode=mode, *args, **kwargs)
         self.s3_client = S3OrcidClient()
+        self.since = since
 
     def _fetch_orcid_data(self, orcid_to_sync, bucket):
         """Fetches a single ORCiD record from S3."""
@@ -60,10 +62,10 @@ class OrcidDataSyncReader(BaseReader):
         """
         date_format = "YYYY-MM-DD HH:mm:ss.SSSSSS"
         date_format_no_millis = "YYYY-MM-DD HH:mm:ss"
-
-        last_sync = arrow.now().shift(
-            days=-current_app.config["VOCABULARIES_ORCID_SYNC_DAYS"]
-        )
+        time_shift = current_app.config["VOCABULARIES_ORCID_SYNC_SINCE"]
+        if self.since:
+            time_shift = self.since
+        last_sync = arrow.now() - timedelta(**time_shift)
 
         file_content = fileobj.read().decode("utf-8")
 
@@ -213,6 +215,32 @@ VOCABULARIES_DATASTREAM_WRITERS = {
 
 
 DATASTREAM_CONFIG = {
+    "readers": [
+        {
+            "type": "tar",
+            "args": {
+                "regex": "\\.xml$",
+            },
+        },
+        {"type": "xml"},
+    ],
+    "transformers": [{"type": "orcid"}],
+    "writers": [
+        {
+            "type": "names-service",
+            "args": {
+                "identity": system_identity,
+            },
+        }
+    ],
+}
+"""ORCiD Data Stream configuration.
+
+An origin is required for the reader.
+"""
+
+# TODO: Used on the jobs and should be set as a "PRESET" (naming to be defined)
+ORCID_PRESET_DATASTREAM_CONFIG = {
     "readers": [
         {
             "type": "orcid-data-sync",
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -8,13 +8,114 @@
 
 """Names datastreams, transformers, writers and readers."""
 
-from invenio_access.permissions import system_identity
+import io
+import tarfile
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import datetime, timedelta
+
+import s3fs
+from flask import current_app
 from invenio_records.dictutils import dict_lookup
 
 from ...datastreams.errors import TransformerError
-from ...datastreams.readers import SimpleHTTPReader
+from ...datastreams.readers import BaseReader, SimpleHTTPReader
 from ...datastreams.transformers import BaseTransformer
 from ...datastreams.writers import ServiceWriter
+
+
+class OrcidDataSyncReader(BaseReader):
+    """ORCiD Data Sync Reader."""
+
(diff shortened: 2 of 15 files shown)


invenio_vocabularies/config.py
```diff
@@ -155,3 +155,14 @@
     "sort": ["name", "count"],
 }
 """Vocabulary type search configuration."""
+
+VOCABULARIES_ORCID_ACCESS_KEY = "TODO"
```
zzacharo, line 159: Shouldn't this config be under "contrib"?

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+        )
+
+        file_content = fileobj.read().decode("utf-8")
+
+        for line in file_content.splitlines()[1:]:  # Skip the header line
+            elements = line.split(",")
+            orcid = elements[0]
+
+            # Lambda file is ordered by last modified date
+            last_modified_str = elements[3]
```
zzacharo, line 65: in the docstring you mention 3 elements while here you are accessing a 4th....is the docstring about the fileObj format outdated?

invenio_vocabularies/config.py
jrcastro2: That would imply to change how we initialize the config, see here: https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/ext.py#L70-L74

It's only getting the main config. To me the different contrib configs look more like static configs, not meant to be used/overriden at the app level.

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+                last_modified_date = datetime.strptime(last_modified_str, date_format)
+            except ValueError:
+                last_modified_date = datetime.strptime(
+                    last_modified_str, date_format_no_millis
+                )
+
+            if last_modified_date >= last_sync:
+                yield orcid
+            else:
+                break
```
kpsherva, line 76: nit: tiny readability improvement:
```suggestion
            if last_modified_date <= last_sync:
                break
            yield orcid
```

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+        file_content = fileobj.read().decode("utf-8")
+
+        for line in file_content.splitlines()[1:]:  # Skip the header line
+            elements = line.split(",")
+            orcid = elements[0]
+
+            # Lambda file is ordered by last modified date
+            last_modified_str = elements[3]
+            try:
+                last_modified_date = datetime.strptime(last_modified_str, date_format)
```
kpsherva, line 67: isn't it more readable and handy to use arrow for date conversions and operations? I think it was our general guildeline

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+        date_format = "%Y-%m-%d %H:%M:%S.%f"
+        date_format_no_millis = "%Y-%m-%d %H:%M:%S"
+
+        last_sync = datetime.now() - timedelta(
+            days=current_app.config["VOCABULARIES_ORCID_SYNC_DAYS"]
+        )
+
+        file_content = fileobj.read().decode("utf-8")
+
+        for line in file_content.splitlines()[1:]:  # Skip the header line
```
kpsherva, line 60: isn't the file simply csv? why not use CSVReader ? it gives you skip header and line.split functions out of the box

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+                for orcid in orcids
+            ]
+            for future in as_completed(futures):
+                result = future.result()
+                if result is not None:
+                    yield result
+
+    def read(self, item=None, *args, **kwargs):
+        """Streams the ORCiD lambda file, process it to get the ORCiDS to sync and yields it's data."""
+        fs = s3fs.S3FileSystem(
```
kpsherva, line 99: shouldn't we encapsulate this in some kind of a "client" class?

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,116 @@
...
+            key=current_app.config["VOCABULARIES_ORCID_ACCESS_KEY"],
+            secret=current_app.config["VOCABULARIES_ORCID_SECRET_KEY"],
+        )
+        # Read the file from S3
+        with fs.open("s3://orcid-lambda-file/last_modified.csv.tar", "rb") as f:
```
kpsherva, line 104: I am thinking if we should make it configurable? also, it would make sense also to be in a "client" class, normally I wouldn't expect the low level file access to be implemented here, but I would also wait for others to share their opinions

invenio_vocabularies/datastreams/writers.py
```diff
@@ -36,6 +41,15 @@ def write(self, stream_entry, *args, **kwargs):
         """
         pass
 
+    def write_many(self, stream_entries, *args, **kwargs):
```
kpsherva, line 44: shouldn't this be abstractmethod?

setup.cfg
```diff
@@ -32,6 +32,7 @@ install_requires =
     invenio-administration>=2.0.0,<3.0.0
     lxml>=4.5.0
     PyYAML>=5.4.1
+    s3fs>=2024.6.1
```
kpsherva, line 35: this should be installed as an extra, otherwise everyone who installs invenio-vocabularies gets also s3fs, which is not desired by default.
In order to make it an extra you will need to make sure the module works also without this package installed. Look at the oaipmh below for inspiration

tests/test_factory.py
```diff
@@ -35,7 +35,7 @@
     "conf, ds_config, service_type",
     [
         (AwardsVocabularyConfig(), awards_ds_config, "awards-service"),
-        (NamesVocabularyConfig(), names_ds_config, "names-service"),
+        (NamesVocabularyConfig(), names_ds_config, "async"),
```
kpsherva, line 38: what are testing here for? not sure I understand what's happening

tests/test_cli.py
```diff
@@ -98,25 +99,67 @@ def names_tar_file(name_xml):
...
+                },
+            }
+        ],
+    }
+
+
+def test_process(app, names_tar_file, mock_config):
+    with patch(
+        "invenio_vocabularies.factories.NamesVocabularyConfig"
+    ) as MockNamesVocabularyConfig:
```
slint, line 130: minor: instead of patching, shouldn't we test the new config directly in datastreams? In general, I see the CLI API as an end-user utility, while the reader/writer/transformer logic should be something that we can unit-test.

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: I think it would be more complex if start to concatenate readers vs having all the logic in the same place. I would be in favour of using the CSVReader if there was no more logic on it but we require to stop reading at a desire time (set in the config), therefore we would need to either create a specific CSVReader for this or update the existing class to support this. And after executing this CSVReader we would need to concatenate again another reader to go for the individual ORCiD files of each person, at least we would need to concatenate 3 readers where - at least - 2 of them would be ORCiD specific.

tests/test_factory.py
jrcastro2: It's checking for the expected writer, now changed to the async writer

tests/test_cli.py
jrcastro2: Hmm I mocked it, because I want to avoid loading 1 day of ORCID data, streaming it from the bucket, which might be quite time consuming, so I simply keep the "original" readers, but we still test the ORCiD transformer and writer, the only thing we don't test is the ORCiD reader, let me know if you think we should change it anyway.

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: Done

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -8,15 +8,121 @@
...
+    def __init__(self, origin=None, mode="r", *args, **kwargs):
+        """Constructor.
+
+        :param origin: Data source (e.g. filepath).
+                       Can be none in case of piped readers.
+        """
+        super().__init__(origin=origin, mode=mode, *args, **kwargs)
+        self.s3_client = S3Client(
+            current_app.config["VOCABULARIES_ORCID_ACCESS_KEY"],
+            current_app.config["VOCABULARIES_ORCID_SECRET_KEY"],
```
kpsherva, line 40: I think this should also be moved to the client

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: DIscussed IRL, the idea is to use python csv lib

invenio_vocabularies/config.py
```diff
@@ -164,5 +164,14 @@
...
-"""ORCID number of days to sync."""
+VOCABULARIES_ORCID_SYNC_SINCE = {
+    "years": 0,
+    "months": 0,
+    "weeks": 0,
+    "days": 1,
+    "hours": 0,
+    "minutes": 0,
+    "seconds": 0,
+    "microseconds": 0,
```
slint, line 175:
```suggestion
    "days": 1,
```

nit: I think we can do with just the one we need, and just write in the docstring that it's parameters passed to `timedelta` constructor.

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -60,9 +61,11 @@ def _process_lambda_file(self, fileobj):
         """
         date_format = "YYYY-MM-DD HH:mm:ss.SSSSSS"
         date_format_no_millis = "YYYY-MM-DD HH:mm:ss"
-
+        time_shift = current_app.config["VOCABULARIES_ORCID_SYNC_SINCE"]
+        if self.since:
+            time_shift = self.since
         last_sync = arrow.now().shift(
-            days=-current_app.config["VOCABULARIES_ORCID_SYNC_DAYS"]
+            **{key: -value for key, value in time_shift.items()}
```
slint, line 68: Instead of `arrow.shift()`, could we use something like:


```python
since = datetime.utcnow() - timedelta(**current_app.config["VOCABULARIES_ORCID_SYNC_SINCE"])
```

slint (approved):
LGTM, some minor comments how to handle the delta

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: I was using this at the beginning and I don't have a personal preference, but there was a suggestion above that we agreed to use arrow. Following the suggestion I have updated it, see comment here: https://github.com/inveniosoftware/invenio-vocabularies/pull/366#discussion_r1686392070

invenio_vocabularies/config.py
jrcastro2: Ok, makes sense 👍

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: Updated! We can use arrow and timedelta, avoiding shift. Thanks!

tests/test_cli.py
slint: Isn't this resolved now that we defaulted back to the original datastream config?

invenio_vocabularies/contrib/names/s3client.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class S3Client:
+    """S3 client."""
+
+    def __init__(self, access_key, secret_key):
+        """Constructor."""
+        try:
+            self.fs = s3fs.S3FileSystem(key=access_key, secret=secret_key)
+        except AttributeError:
```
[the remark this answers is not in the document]
jrcastro2: Good point!

PR #370, opened 2024-07-24
contrib: names: update field boosts and mapping

sakshamarora1:
Fixes: https://github.com/inveniosoftware/invenio-app-rdm/issues/2756

The names in the database and index are inverted, therefore the suggest API first tries to search on the last name and prioritizes that due to how edge ngram works for the suggest API.
This solution add suggest API for given name for better results.

# ATTENTION

tests fail due to https://github.com/inveniosoftware/invenio-vocabularies/pull/372, to be rebased and merged after #372

### DIFF
--- a/invenio_vocabularies/contrib/names/config.py
+++ b/invenio_vocabularies/contrib/names/config.py
@@ -28,12 +28,16 @@ class NamesSearchOptions(SearchOptions):
 
     suggest_parser_cls = SuggestQueryParser.factory(
         fields=[
-            "name^100",
-            "family_name^100",
+            "name.suggest^80",
+            "name^80",
+            "family_name^70",
+            "given_name.suggest^100",
             "given_name^100",
             "identifiers.identifier^20",
             "affiliations.name^10",
-        ],
+        ],  # type has to be bool_prefix to use suggest._index_prefix field during querying
+        operator="AND",
+        fuzziness="AUTO",
     )
 
     sort_default = "bestmatch"
--- a/invenio_vocabularies/contrib/names/config.py
+++ b/invenio_vocabularies/contrib/names/config.py
@@ -28,15 +28,13 @@ class NamesSearchOptions(SearchOptions):
 
     suggest_parser_cls = SuggestQueryParser.factory(
         fields=[
-            "name.suggest^80",
-            "name^80",
-            "family_name^70",
-            "given_name.suggest^100",
             "given_name^100",
+            "name^70",
+            "family_name^50",
             "identifiers.identifier^20",
-            "affiliations.name^10",
-        ],  # type has to be bool_prefix to use suggest._index_prefix field during querying
-        operator="AND",
+            "affiliations.name^20",
+        ],
+        type="most_fields",  # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#multi-match-types
         fuzziness="AUTO",
     )
 
--- a/invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
+++ b/invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
@@ -125,6 +125,16 @@
           "char_filter": ["strip_special_chars"],
           "filter": ["lowercase", "asciifolding"]
         }
+      },
+      "filter": {
+        "lowercase": {
+          "type": "lowercase",
+          "preserve_original": true
+        },
+        "asciifolding": {
+          "type": "asciifolding",
+          "preserve_original": true
+        }
       }
     }
   }
--- a/invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
+++ b/invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
@@ -9,11 +9,34 @@
         }
       },
       "analyzer": {
+        "accent_edge_analyzer": {
+          "tokenizer": "standard",
+          "type": "custom",
+          "char_filter": ["strip_special_chars"],
+          "filter": [
+            "lowercase",
+            "asciifolding",
+            "edgegrams"
+          ]
+        },
         "accent_analyzer": {
           "tokenizer": "standard",
           "type": "custom",
           "char_filter": ["strip_special_chars"],
-          "filter": ["lowercase", "asciifolding"]
+          "filter": [
+            "lowercase",
+            "asciifolding"
+          ]
+        }
+      },
+      "normalizer": {
+        "accent_normalizer": {
+          "type": "custom",
+          "char_filter": ["strip_special_chars"],
+          "filter": [
+            "lowercase",
+            "asciifolding"
+          ]
         }
       },
       "filter": {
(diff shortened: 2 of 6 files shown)


invenio_vocabularies/contrib/names/config.py
```diff
@@ -28,9 +28,9 @@ class NamesSearchOptions(SearchOptions):
 
     suggest_parser_cls = SuggestQueryParser.factory(
         fields=[
-            "name^100",
-            "family_name^100",
-            "given_name^100",
+            "name^80",
```
zzacharo, line 31: shouldn't these also utilize the `search_as_you_type` suggest field as well?

invenio_vocabularies/contrib/names/jsonschemas/names/name-v2.0.0.json
```diff
@@ -0,0 +1,45 @@
+{
```
jrcastro2, line 1: Same comment as for the records mapping, if this is not needed to update I would avoid updating only the version. To be checked.

invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
```diff
@@ -0,0 +1,106 @@
...
+      "given_name": {
+        "type": "text",
+        "fields": {
+          "suggest": {
+            "type": "search_as_you_type"
+          }
+        }
+      },
+      "family_name": {
+        "type": "text"
```
jrcastro2, line 51: Comment Do we want to add it as well to the family_name?

jrcastro2 (commented):
LGTM! Just a couple of comments.

invenio_vocabularies/contrib/names/config.py
sakshamarora1: Yes, added, thanks!

invenio_vocabularies/contrib/names/jsonschemas/names/name-v2.0.0.json
sakshamarora1: Yes, not needed, removed!

invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
sakshamarora1: It is not needed since the `name` field already starts from the last name

carlinmack:
can you add a test like https://github.com/inveniosoftware/invenio-vocabularies/blob/master/tests/contrib/funders/test_funders_resource.py#L124 :)

tests/contrib/names/test_names_resource.py
```diff
@@ -85,3 +144,26 @@ def test_names_resolve(client, example_name, h, prefix):
...
+    res = client.get(f"{prefix}?suggest=saksham", headers=h)
+    assert res.status_code == 200
+    assert res.json["hits"]["total"] == 1
+    assert res.json["hits"]["hits"][0]["name"] == "Arœra, Säksham"
+
+    # With incomplete
+    res = client.get(f"{prefix}?suggest=jos", headers=h)  # jos
+    assert res.status_code == 200
+    assert res.json["hits"]["total"] == 1
+    assert res.json["hits"]["hits"][0]["name"] == "Gonzalez Lopez, Jose Benito"
```
jrcastro2, line 169: Suggestion: Often we use fake names for testing ;)

jrcastro2:
ES tests are failing

sakshamarora1:
@jrcastro2 ES tests are removed in #372 😅
Didn't wanna duplicate it here to prevent merge conflicts

invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
```diff
@@ -0,0 +1,141 @@
...
+            "type": "keyword"
+          },
+          "status": {
+            "type": "keyword"
+          }
+        }
+      }
+    }
+  },
+  "settings": {
```
kpsherva, line 112: nit: could you move it on top of the file? everywhere else we keep the settings on top and it would be better to be consistent - people will expect to find it on top

kpsherva (commented):
looks, good, needs the tests to pass

invenio_vocabularies/contrib/names/mappings/os-v2/names/name-v2.0.0.json
```diff
@@ -0,0 +1,141 @@
...
+          },
+          "id": {
+            "type": "keyword"
+          },
+          "name": {
+            "type": "keyword",
+            "fields": {
+              "suggest": {
+                "type": "search_as_you_type"
+              }
```
slint, line 117: - Let's change `type` to `text`
- Let's add the `accent_analyzer` to the `name.suggest` sub-field

invenio_vocabularies/contrib/names/config.py
```diff
@@ -28,12 +28,16 @@ class NamesSearchOptions(SearchOptions):
...
         fields=[
-            "name^100",
-            "family_name^100",
+            "name.suggest^80",
+            "name^80",
+            "family_name^70",
+            "given_name.suggest^100",
             "given_name^100",
             "identifiers.identifier^20",
             "affiliations.name^10",
```
slint, line 37: - For `affiliations.name`, we should instead use `affiliations.name.suggest`
- nit: reorder the fields by "boosting weight"

invenio_vocabularies/contrib/names/mappings/os-v2/names/name-v2.0.0.json
```diff
@@ -0,0 +1,141 @@
...
+      "tags": {
+        "type": "keyword"
+      },
+      "name_sort": {
+        "type": "keyword"
+      },
+      "name": {
+        "type": "text",
+        "analyzer": "accent_analyzer",
+        "search_analyzer": "accent_analyzer",
```
slint, line 65: Since we're only doing "autocomplete"-style search for names, we don't have to use the analyzer on the "base" text field.

sakshamarora1: So the `search_as_you_type` fields are optimized in such a way that fuzziness cannot be applied on the fields generated by them (`._index_prefix` in this case). So, not removing the analyzer from the base text field for fuzziness functionality to work.

In the recent commit, I have updated the mappings to use edge ngrams and then doing search on the generated tokens instead of search as you type so that fuzziness can be applied.
Looks like dropping the extra sub fields in favor of having ngram tokens is taking up around the same storage with similar search speed.

tests/contrib/names/test_names_resource.py
```diff
@@ -85,3 +144,33 @@ def test_names_resolve(client, example_name, h, prefix):
...
+    res = client.get(f"{prefix}?suggest=davisson", headers=h)
+    assert res.status_code == 200
+    assert res.json["hits"]["total"] == 1
+    assert res.json["hits"]["hits"][0]["name"] == "Davidson, Niels John"
+
+    # With accent
+    res = client.get(f"{prefix}?suggest=dwayne", headers=h)
+    assert res.status_code == 200
+    assert res.json["hits"]["total"] == 1
+    assert res.json["hits"]["hits"][0]["name"] == "Johnsœn, Dwäyne"
```
slint, line 162: Can you also add a test for `?suggest=Dwäyne`, i.e capitalized and accent? I guess it should return the same results.

zzacharo (commented):
@sakshamarora1 there is only one comment on this from @slint. Otherwise, it seems good to go!

PR #376, opened 2024-08-02
names: validate entry full names

yashlamba:


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -140,6 +141,11 @@ def __init__(self, *args, test_mode=True, **kwargs):
 class OrcidTransformer(BaseTransformer):
     """Transforms an ORCiD record into a names record."""
 
+    def _is_valid_name(self, name):
+        re.DEFAULT_VERSION = re.VERSION1
+        regex = r"[\p{P}\p{S}\p{Nd}\p{No}\p{Emoji}--,.()']"
+        return not bool(re.search(regex, name, re.UNICODE))
```
slint, line 147:
```suggestion
        return not bool(re.search(regex, name, (re.UNICODE | re.VERSION1)))
```

To avoid globally modifying the `regex` library

PR #380, opened 2024-08-08
subjects: switch to datastreams

### LINKED ISSUE #322: Switch subjects to datastreams
- subject vocabularies is still a fixture (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/vocabularies.py)
- There is a bunch of itereators (`YamlIterator`...) which I assume are like datastream readers
-  I guess Schemes (`VocabularyEntryWithSchemes`) are needed to differentiate between different sources for one specific vocabulary. So for example you'd load into `subjects` from GND _and_ MeSH and so on. 
    - I'm mot sure how this would be ported in terms of datastreams

yashlamba:
Continuation of: #332 

Closes #322


ptamarit (approved):
OK for me expect the 2 TODOs which should be removed.

ptamarit (approved):

ptamarit:
Just rebased from `master`, to include `tests: remove unecessary "enzyme-to-json" package` (7e992f2ad25d20640a6b9a69f08a4c0161a6c671), in order to fix the JS Tests in this pull request.

PR #372, opened 2024-07-24
mappings: add edge_ngram prefix search

carlinmack:
:heart: Thank you for your contribution!

Close https://github.com/inveniosoftware/invenio-app-rdm/issues/2759
Close https://github.com/inveniosoftware/invenio-app-rdm/issues/2761

### Description

> Please describe briefly your pull request.

for the three vocabularies (funders, affiliations and awards) I created a test for searching for the prefix, which failed. I then added the edge_ngram search to the mapping which made the test pass

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

PR #385, opened 2024-08-14
subjects: Add MeSH datastream

yashlamba:
Closes https://github.com/zenodo/zenodo-rdm/issues/926


invenio_vocabularies/contrib/subjects/mesh/datastreams.py
```diff
@@ -0,0 +1,43 @@
...
+        entry_data = stream_entry.entry
+
+        # ID in MeSH data is the URL, ex. https://id.nlm.nih.gov/mesh/D000001
+        # We just want to use the ID prefixed by "mesh:""
+        try:
+            mesh_id = entry_data["id"].split("/")[-1]
+        except Exception:
+            raise TransformerError("Not a valid MeSH ID.")
+
+        entry_data["id"] = "mesh:" + mesh_id
```
slint, line 32:
```suggestion
        entry_data["id"] = f"mesh:{mesh_id}"
```

nit: f-strings is short for `ftw strings` 🧠

invenio_vocabularies/config.py
```diff
@@ -104,14 +105,19 @@
 }
 """Names allowed identifier schemes."""
 
-# configure CUSTOM_VOCABULARY_TYPES to differentiate output. Is used in VocabulariesServiceConfig
+VOCABULARIES_SUBJECTS_SCHEMES = {
+    "gnd": {"label": _("GND"), "validator": idutils.is_gnd, "datacite": "GND"},
```
slint, line 109: minor: we should also have a `mesh` scheme, though it has to be implemented in `idutils`. It would just be `mesh:D000001`, but it would allow to store a more compact version of the identifier instead of the full URL, and generate the fully resolvable URL through it

invenio_vocabularies/config.py
```diff
@@ -104,14 +105,19 @@
...
+"""Subjects allowed identifier schemes."""
+
 VOCABULARIES_CUSTOM_VOCABULARY_TYPES = [
     "names",
     "affiliations",
     "awards",
     "funders",
     "subjects",
 ]
+"""Configure CUSTOM_VOCABULARY_TYPES to differentiate output. Is used in VocabulariesServiceConfig."""
```
slint, line 120:
```suggestion
"""List of custom vocabulary types."""
```

invenio_vocabularies/contrib/subjects/config.py
```diff
@@ -40,7 +46,7 @@ class SubjectsSearchOptions(SearchOptions):
         ),
         "subject": dict(
             title=_("Name"),
-            fields=["subject_sort"],
+            fields=["indexed_at"],
```
slint, line 49: minor: we should keep `subject_sort` as the sorting field (it's still present in the schema and is a required field).

invenio_vocabularies/contrib/subjects/config.py
```diff
@@ -2,30 +2,36 @@
...
 class SubjectsSearchOptions(SearchOptions):
     """Search options."""
 
     suggest_parser_cls = FilteredSuggestQueryParser.factory(
         filter_field="scheme",
         fields=[  # suggest fields
-            "subject^100",
-            "subject._2gram",
-            "subject._3gram",
+            "title.*^100",
```
slint, line 33:
```suggestion
            "subject",
            "title.*^100",
```

- We can add back the `subject` field
- Regarding the localized titles, see issue and comment at https://github.com/inveniosoftware/invenio-records-resources/pull/594#issuecomment-2288980884 on how this should be handled.

invenio_vocabularies/contrib/subjects/datastreams.py
```diff
@@ -0,0 +1,55 @@
...
+    """Subjects Service Writer."""
+
+    def __init__(self, *args, **kwargs):
+        """Constructor."""
+        service_or_name = kwargs.pop("service_or_name", "subjects")
+        super().__init__(service_or_name=service_or_name, *args, **kwargs)
+
+    def _entry_id(self, entry):
+        """Get the id from an entry."""
+        return entry["id"]
```
slint, line 30: comment: if this subclass only exists because of this part, we should have a look at how to make this generic/easier to configure, since just subclassing for this reasons is very inefficient

invenio_vocabularies/contrib/subjects/mappings/os-v1/subjects/subject-v1.0.0.json
```diff
@@ -1,5 +1,16 @@
 {
   "mappings": {
+    "dynamic_templates": [
+      {
+        "i18n_title": {
+          "path_match": "title.*",
+          "match_mapping_type": "string",
+          "mapping": {
+            "type": "search_as_you_type"
+          }
```
slint, line 10: minor/shelve: we should apply the same search optimizations that are available at https://github.com/inveniosoftware/invenio-vocabularies/pull/372. Can be done in a separate PR

invenio_vocabularies/contrib/subjects/schema.py
```diff
@@ -24,6 +33,16 @@ class SubjectSchema(BaseVocabularySchema):
     id = SanitizedUnicode(required=True)
     scheme = SanitizedUnicode(required=True)
     subject = SanitizedUnicode(required=True)
+    title = i18n_strings
+    synonyms = fields.List(SanitizedUnicode())
+
+    @pre_load
+    def add_subject_from_title(self, data, **kwargs):
+        """Add subject from title if not present."""
+        locale = current_app.config.get("BABEL_DEFAULT_LOCALE", "en")
```
slint, line 42:
```suggestion
        locale = get_locale()
```

minor: this is more reliable in terms of how to fetch the default locale. If this fails for app/request context reasons, it should be fixed upstream so that it doesn't.

slint (commented):
Some minor comments. All the ones relevant to search/mappings are shelveable and can be addressed in a separate PR.

invenio_vocabularies/contrib/subjects/datastreams.py
yashlamba: It confused me as well, but I guess updating it might need a major? (but it's a "private" function) so maybe not really?

I think all vocab writers can use this change. Additionally, this part is also funny to me: https://github.com/inveniosoftware/invenio-vocabularies/blob/570f268414b1aece6dc60c1b274e4de29d873f0c/invenio_vocabularies/contrib/affiliations/datastreams.py#L22-L25

to register just a specific vocab writer, we do this, I would just prefer using the service writer with the name/just fix the name and don't care about the arg.

yashlamba:
Subjects related suggestions resolved here: https://github.com/inveniosoftware/invenio-vocabularies/pull/380/commits/b73c96919a960c15f934454f20a36281cfef13b4

PR #388, opened 2024-08-18
ror: fix dump modified date comparison

slint:
Requires https://github.com/inveniosoftware/invenio-rdm-records/pull/1798 to be merged and deployed on Zenodo (so that the JSON-LD export format includes the `dateCreated` field)


yashlamba (approved):

slint:
Rebased, but will probably have to cherry-pick and release as a patch under a `maint-4.x` branch, since we're planning to release the search mappings as a a major bump.

PR #389, opened 2024-08-21
ci: run on maint/feature branches

yashlamba:


.github/workflows/tests.yml
```diff
@@ -13,6 +13,8 @@ on:
   push:
     branches:
       - master
+      - "feature/*"
```
slint, line 16: I would also add it on `pull_request`, since we sometimes develop and open PRs against the `features/*`

PR #396, opened 2024-09-02
resource: add permission check to names search

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/191

### DIFF
--- a/invenio_vocabularies/contrib/names/names.py
+++ b/invenio_vocabularies/contrib/names/names.py
@@ -21,7 +21,8 @@ from invenio_records_resources.records.systemfields import (
 )
 from invenio_records_resources.resources.records.headers import etag_headers
 
-from ...services.permissions import PermissionPolicy
+from invenio_vocabularies.contrib.names.permissions import NamesPermissionPolicy
+
 from ..affiliations.api import Affiliation
 from .config import NamesSearchOptions, service_components
 from .schema import NameSchema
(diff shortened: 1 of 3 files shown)


alejandromumo (commented):
LGTM, I am pinging Alex for a double check that the endpoint can be also protected by login on Zenodo

WDYT @slint

invenio_vocabularies/contrib/names/resources.py
```diff
@@ -52,3 +61,40 @@ def name_resolve_by_id(self):
         )
 
         return item.to_dict(), 200
+
+    @request_extra_args
+    @request_search_args
+    @response_handler(many=True)
+    @login_required
```
slint, line 68: minor: shouldn't we instead be using the permission policy and the `AuthenticatedUser` need? If this is how this is done in other places, feel free to ignore :)

slint:
As discussed IRL, the change is fine, especially for the `names` vocabulary since it's only used in forms that require authentication anyways.

invenio_vocabularies/contrib/names/resources.py
jrcastro2: That's a good point! 
It really depends on whether we want to protect just the endpoint, as the login_required decorator does, or if we also want to secure the service methods. I believe we should go with the latter approach, so I’ll make the necessary changes based on your suggestion. Thanks!

slint (approved):

PR #398, opened 2024-09-06
Add non-searchable tags

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/193
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/194

### DIFF
--- a/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
+++ b/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
@@ -24,6 +24,14 @@
       "items": {"$ref": "local://definitions-v1.0.0.json#/identifiers_with_scheme"},
       "uniqueItems": true
     },
+    "props": {
+      "type": "object",
+      "patternProperties": {
+        "^.*$": {
+          "type": "string"
+        }
+      }
+    },
     "affiliations": {
       "description": "Affiliations of the person.",
       "type": "array",
--- a/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
+++ b/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
@@ -15,20 +15,35 @@
       "description": "Identifier of the name scheme.",
       "$ref": "local://definitions-v1.0.0.json#/identifier"
     },
-    "name": {"type": "string"},
-    "given_name": {"type": "string"},
-    "family_name": {"type": "string"},
+    "name": {
+      "type": "string"
+    },
+    "given_name": {
+      "type": "string"
+    },
+    "family_name": {
+      "type": "string"
+    },
     "identifiers": {
       "description": "Identifiers for the person.",
       "type": "array",
-      "items": {"$ref": "local://definitions-v1.0.0.json#/identifiers_with_scheme"},
+      "items": {
+        "$ref": "local://definitions-v1.0.0.json#/identifiers_with_scheme"
+      },
       "uniqueItems": true
     },
     "props": {
       "type": "object",
       "patternProperties": {
         "^.*$": {
-          "type": "string"
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "boolean"
+            }
+          ]
         }
       }
     },
--- a/invenio_vocabularies/contrib/names/permissions.py
+++ b/invenio_vocabularies/contrib/names/permissions.py
@@ -10,11 +10,23 @@
 
 from invenio_records_permissions.generators import AuthenticatedUser, SystemProcess
 
-from ...services.permissions import PermissionPolicy
+from invenio_vocabularies.services.generators import IfTags
+from invenio_vocabularies.services.permissions import PermissionPolicy
 
 
 class NamesPermissionPolicy(PermissionPolicy):
-    """Permission policy."""
+    """Names permission policy.
 
-    can_search = [SystemProcess(), AuthenticatedUser()]
-    can_read = [SystemProcess(), AuthenticatedUser()]
+    Names endpoints are protected, only authenticated users can access them.
+    """
+
+    can_search = [
+        SystemProcess(),
+        IfTags(exclude=["unlisted"], only_authenticated=True),
+    ]
+    can_read = [SystemProcess(), IfTags(exclude=["unlisted"], only_authenticated=True)]
+    # this permission is needed for the /api/vocabularies/ endpoint
+    can_list_vocabularies = [
+        SystemProcess(),
+        IfTags(exclude=["unlisted"], only_authenticated=True),
+    ]
--- a/invenio_vocabularies/contrib/names/permissions.py
+++ b/invenio_vocabularies/contrib/names/permissions.py
@@ -22,11 +22,9 @@ class NamesPermissionPolicy(PermissionPolicy):
 
     can_search = [
         SystemProcess(),
-        IfTags(exclude=["unlisted"], only_authenticated=True),
+        AuthenticatedUser(),
     ]
-    can_read = [SystemProcess(), IfTags(exclude=["unlisted"], only_authenticated=True)]
-    # this permission is needed for the /api/vocabularies/ endpoint
-    can_list_vocabularies = [
+    can_read = [
         SystemProcess(),
-        IfTags(exclude=["unlisted"], only_authenticated=True),
+        IfTags(["unlisted"], then_=[SystemProcess()], else_=[AuthenticatedUser()]),
     ]
--- a/invenio_vocabularies/contrib/names/services.py
+++ b/invenio_vocabularies/contrib/names/services.py
@@ -19,11 +19,12 @@ NamesServiceConfig = record_type.service_config_cls
 class NamesService(record_type.service_cls):
     """Name service."""
 
-    def resolve(self, identity, id_, id_type):
+    def resolve(self, identity, id_, id_type, many=False):
         """Get the record with a given identifier.
 
-        This method assumes that the are no duplicates in the system
-        (i.e. only one name record can have a pair of identifier:scheme).
+        param id_: The identifier value.
+        param id_type: The identifier type.
+        param many: If True, return a list of records.
         """
         search_query = dsl.Q(
             "bool",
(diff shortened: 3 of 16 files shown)


invenio_vocabularies/contrib/names/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+    can_search = [
+        SystemProcess(),
+        Tags(exclude=["non-searchable"], only_authenticated=True),
+    ]
+    can_read = [SystemProcess(), AuthenticatedUser()]
+    # this permission is needed for the /api/vocabularies/ endpoint
+    can_list_vocabularies = [
+        SystemProcess(),
+        Tags(exclude=["non-searchable"], only_authenticated=True),
```
kpsherva, line 28: side question, for me to understand how this works. How can I know what tags are available?

jrcastro2: Hmm, I think it's only documented in the RFC: https://github.com/inveniosoftware/rfcs/blob/fdb09b8b86263607364c8cc554b3a6580ccba2e2/rfcs/rdm-0077-vocabulary-harvesting.md#data-model

There is PR adding the `non-searchable`explanation, however maybe we should document htis better in the docs (?)

invenio_vocabularies/contrib/names/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+from invenio_vocabularies.services.generators import Tags
+from invenio_vocabularies.services.permissions import PermissionPolicy
+
+
+class NamesPermissionPolicy(PermissionPolicy):
+    """Names permission policy."""
+
+    can_search = [
+        SystemProcess(),
+        Tags(exclude=["non-searchable"], only_authenticated=True),
```
ptamarit, line 22: Is non-searchable the best terminology?
Another idea from Nico would be "unlisted"?

invenio_vocabularies/services/generators.py
```diff
@@ -0,0 +1,60 @@
...
+#
+
+"""Vocabulary generators."""
+
+from invenio_access import any_user, authenticated_user
+from invenio_records_permissions.generators import Generator
+from invenio_search.engine import dsl
+
+
+class AnyUser(Generator):
```
ptamarit, line 17: Why this specific `AnyUser` class while it's unused and we already have a general one?

invenio_vocabularies/services/permissions.py
```diff
@@ -9,17 +9,19 @@
...
+from invenio_vocabularies.services.generators import Tags
 
 
 class PermissionPolicy(RecordPermissionPolicy):
     """Permission policy."""
 
-    can_search = [SystemProcess(), AnyUser()]
-    can_read = [SystemProcess(), AnyUser()]
+    can_search = [SystemProcess(), Tags(exclude=["non-searchable"])]
+    can_read = [SystemProcess(), Tags(exclude=["non-searchable"])]
```
ptamarit, line 21: Is it really what we want here that nobody at all can search non-searchable names?

invenio_vocabularies/services/generators.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def query_filter(self, **kwargs):
+        """Match only searchable values in search."""
+        return dsl.Q(
+            "bool",
+            must_not=[dsl.Q("term", tags="non-searchable")],
+        )
+
+
+class Tags(Generator):
```
ptamarit, line 32: Maybe it's about the naming or how it's used, but we spent time reading this class and are still struggling to understand it.

invenio_vocabularies/services/generators.py
```diff
@@ -0,0 +1,60 @@
...
+    """Allows any user."""
```
ntarocco, line 33: Discussed, please add more explanations related to `only_authenticated` and its purpose.

invenio_vocabularies/contrib/names/permissions.py
```diff
@@ -0,0 +1,29 @@
...
+"""Vocabulary permissions."""
+
+from invenio_records_permissions.generators import AuthenticatedUser, SystemProcess
+
+from invenio_vocabularies.services.generators import Tags
+from invenio_vocabularies.services.permissions import PermissionPolicy
+
+
+class NamesPermissionPolicy(PermissionPolicy):
+    """Names permission policy."""
```
ntarocco, line 18: Could you also explain better that the names vocab is now restricted to avoid leaking names?

invenio_vocabularies/services/generators.py
ntarocco: Checking other generators in other modules, maybe we could use `IfTags`

invenio_vocabularies/services/generators.py
ntarocco: As discussed IRL, to be removed.

invenio_vocabularies/services/permissions.py
jrcastro2: People with admin permission can search for it. If there is other permissions that should have access to it let me know. But to me it makes sense that only administrators do have.

invenio_vocabularies/contrib/names/permissions.py
```diff
@@ -10,11 +10,23 @@
...
 
-    can_search = [SystemProcess(), AuthenticatedUser()]
-    can_read = [SystemProcess(), AuthenticatedUser()]
+    Names endpoints are protected, only authenticated users can access them.
+    """
+
+    can_search = [
+        SystemProcess(),
+        IfTags(exclude=["unlisted"], only_authenticated=True),
+    ]
```
slint, line 26:
```suggestion
    can_search = [
        SystemProcess(),
        IfTags(exclude=["unlisted"]),
        AuthenticatedUser(),
    ]
```

minor: wouldn't that be a bit possible in terms of reusing/composing with existing generators? The `only_authenticated` parameter feels a bit ad-hoc

invenio_vocabularies/contrib/names/services.py
```diff
@@ -36,20 +37,25 @@ def resolve(self, identity, id_, id_type):
...
-        self.require_permission(identity, "read", record=record)
-
-        return self.result_item(
-            self,
-            identity,
-            record,
-            links_tpl=self.links_item_tpl,
-        )
+        if many:
+            return self.result_list(self, identity, results)
```
kpsherva, line 51: is there permission needed to read many? I had this thought seeing the permission required below

invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json
```diff
@@ -15,15 +15,41 @@
...
+        "^.*$": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "array"
```
ntarocco, line 47: Is array wanted and maybe copied from somewhere else? It might be difficult to have good search queries with arrays.

jrcastro2: I thought to use to store affiliations at the beginning but not anymore needed as we only store them at the record level not in the props, I will remove it 👍

invenio_vocabularies/contrib/names/services.py
jrcastro2: In general it doesn't seem that we do a permission check for the read_many and read_all func. But I agree with you we should most likely add it, I am adding it for this one.

invenio_vocabularies/contrib/names/permissions.py
jrcastro2: Right! I agree but unfortunately it doesn't work like that, at the moment if we change it to your suggestion (which was my first thought/try as well) it will simply grant access if any of the needs is matched, meaning that ti will match for unlisted OR authenticated, while what we want to achieve is to grant access to unlisted AND authenticated for Names vocab while for the rest we want to grant access to unlisted AND any user.

That's why I added that param. The only easy alternative I can think of is creating 2 generators that would be `AnyUserIfTags` and `AuthenticatedUserIfTags` maybe this would be more readable ... Unless there are better sugesstions, they are more than welcomed!

invenio_vocabularies/services/generators.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, include=None, exclude=None, only_authenticated=False):
+        """Constructor."""
+        self.include = include or []
+        self.exclude = exclude or []
+        self.only_authenticated = only_authenticated
+
+    def needs(self, **kwargs):
+        """Enabling Needs."""
+        return [authenticated_user] if self.only_authenticated else [any_user]
```
ntarocco, line 32: Is only authenticated correct here? wasn'it it that only admins should be able to get the unlisted?

invenio_vocabularies/contrib/names/permissions.py
ntarocco: I also agree that is not idea to "hardcode" the authenticated part in a `IfTags`.
I would rather specialize the unlisted instead, which is a bit different, as only admins should be able to see it. Let's discuss IRL.

tests/contrib/names/test_name_permissions.py
```diff
@@ -0,0 +1,87 @@
...
+    assert service.config.indexer_queue_name == "names"
+    # Create it
+    item = service.create(identity, non_searchable_name_data)
+    id_ = item.id
+
+    # Index document in ES
+    assert indexer.refresh()
+
+    with pytest.raises(PermissionDeniedError):
+        # Read - only searchable values should be returned
```
zzacharo, line 63: Wrong comment? In this case a PermissionDenied is raised no?

invenio_vocabularies/contrib/names/permissions.py
```diff
@@ -10,11 +10,21 @@
...
+    Names endpoints are protected, only authenticated users can access them.
+    """
+
+    can_search = [
+        SystemProcess(),
+        AuthenticatedUser(),
+    ]
+    can_read = [
+        SystemProcess(),
+        IfTags(["unlisted"], then_=[SystemProcess()], else_=[AuthenticatedUser()]),
```
zzacharo, line 29: Sorry this mean that the `system_identity` can only read these records. I see in the test that you mention admins. Who should be able to read these in the end?

tests/conftest.py
```diff
@@ -113,6 +118,17 @@ def identity():
     return i
 
 
+@pytest.fixture(scope="module")
+def superuser_identity():
+    """Super user identity to interact with the services."""
+    i = Identity(2)
+    i.provides.add(UserNeed(2))
+    i.provides.add(any_user)
+    i.provides.add(system_process)
```
zzacharo, line 127: My confusion about who can read the record is essentially this fixture. The superuser can read because they have the `system_process` need not because of the `superuser_access`

invenio_vocabularies/contrib/names/permissions.py
ntarocco: SuperUser (admins) has always access by default. In general, we put SystemProcess to signal that and allow CLI.

invenio_vocabularies/contrib/names/services.py
```diff
@@ -36,20 +37,26 @@ def resolve(self, identity, id_, id_type):
...
-
-        return self.result_item(
-            self,
-            identity,
-            record,
-            links_tpl=self.links_item_tpl,
-        )
+        if many:
+            self.require_permission(identity, "read")
+            return self.result_list(self, identity, results)
```
ntarocco, line 52: Here we don't really check for the unlisted tag. Any possible solution?

invenio_vocabularies/contrib/names/services.py
```diff
@@ -36,20 +37,26 @@ def resolve(self, identity, id_, id_type):
...
-        self.require_permission(identity, "read", record=record)
-
-        return self.result_item(
-            self,
-            identity,
-            record,
-            links_tpl=self.links_item_tpl,
-        )
+        if many:
+            self.require_permission(identity, "read")
```
zzacharo, line 51:
```suggestion
            self.require_permission(identity, "read", record=record)
```
We miss to pass the record no?

invenio_vocabularies/contrib/names/permissions.py
jrcastro2: Discussed IRL and agreed to use a generator like this, change implemented
```python
IfTags(["unlisted"], then_=[SystemProcess()], else_=[AnyUser()])
```

PR #409, opened 2024-10-08
Rename patched filters so the normalizer uses the default ones

### LINKED ISSUE #408: The newly introduced normalizer in the funders vocab generates two tokens when acronyms contain diacritics
## Package version (if known): <version>

Since v5.1.0

## Describe the bug

The normalization for new or updated entries in the funders index leads to the following problem:

The normalization token stream is expected to produce exactly 1 token, but got 2+ for analyzer analyzer name[accent_normalizer], analyzer [org.opensearch.index.analysis.CustomAnalyzer@33ac3672], analysisMode [ALL] and input "FüAkBw"

The document will not be indexed.

## Steps to Reproduce
1. Run `invenio index destroy --yes-i-know`
2. Run `invenio index init`
3. Run `invenio rdm rebuild-all-indices`

## Expected behavior

All records and vocabularies should be reindexed.

wgresshoff:

Close #408 

### Description

The normalizer that was introduced in the funders index with v5.1.0 generates two tokens when used on documents that contain diacritics in the acronym. This behaviour occurs when in the filter definition `"preserve_original": true` is used. This is ok for analyzers, so this PR just renames the filters and takes care the analyzers use the new ones and the normailzer uses the original ones.

### Checklist

- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).

**Frontend**

- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_vocabularies/contrib/funders/mappings/os-v1/funders/funder-v2.0.0.json
+++ b/invenio_vocabularies/contrib/funders/mappings/os-v1/funders/funder-v2.0.0.json
@@ -14,8 +14,8 @@
           "type": "custom",
           "char_filter": ["strip_special_chars"],
           "filter": [
-            "lowercase",
-            "asciifolding",
+            "rdm_lowercase",
+            "rdm_asciifolding",
             "edgegrams"
           ]
         },
@@ -24,8 +24,8 @@
           "type": "custom",
           "char_filter": ["strip_special_chars"],
           "filter": [
-            "lowercase",
-            "asciifolding"
+            "rdm_lowercase",
+            "rdm_asciifolding"
           ]
         }
       },
@@ -40,11 +40,11 @@
         }
       },
       "filter": {
-        "lowercase": {
+        "rdm_lowercase": {
           "type": "lowercase",
           "preserve_original": true
         },
-        "asciifolding": {
+        "rdm_asciifolding": {
           "type": "asciifolding",
           "preserve_original": true
         },
--- a/invenio_vocabularies/contrib/funders/mappings/os-v1/funders/funder-v2.0.0.json
+++ b/invenio_vocabularies/contrib/funders/mappings/os-v1/funders/funder-v2.0.0.json
@@ -14,8 +14,8 @@
           "type": "custom",
           "char_filter": ["strip_special_chars"],
           "filter": [
-            "rdm_lowercase",
-            "rdm_asciifolding",
+            "lowercasepreserveoriginal",
+            "asciifoldingpreserveoriginal",
             "edgegrams"
           ]
         },
@@ -24,8 +24,8 @@
           "type": "custom",
           "char_filter": ["strip_special_chars"],
           "filter": [
-            "rdm_lowercase",
-            "rdm_asciifolding"
+            "lowercasepreserveoriginal",
+            "asciifoldingpreserveoriginal"
           ]
         }
       },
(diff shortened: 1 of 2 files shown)


invenio_vocabularies/contrib/funders/mappings/os-v2/funders/funder-v2.0.0.json
```diff
@@ -40,11 +40,11 @@
         }
       },
       "filter": {
-        "lowercase": {
+        "rdm_lowercase": {
           "type": "lowercase",
           "preserve_original": true
         },
-        "asciifolding": {
+        "rdm_asciifolding": {
```
sakshamarora1, line 47: Hello! Sorry for the nitpick: We are trying to keep the settings same across all mappings so can we keep the name [asciifoldingpreserveoriginal](https://github.com/inveniosoftware/invenio-vocabularies/blob/9f69aa7a46da9c1168712157a0189878288f7c0a/invenio_vocabularies/contrib/names/mappings/os-v2/names/name-v2.0.0.json#L47)? Same for lowercasepreserveoriginal.

 However, the change looks good to go!

sakshamarora1 (approved):

invenio_vocabularies/contrib/funders/mappings/os-v2/funders/funder-v2.0.0.json
wgresshoff: Hi, yes sure, I wasn't aware there was a solution elsewhere. I'll change the filter names.

invenio_vocabularies/contrib/funders/mappings/os-v1/funders/funder-v2.0.0.json
```diff
@@ -14,8 +14,8 @@
           "type": "custom",
           "char_filter": ["strip_special_chars"],
           "filter": [
-            "lowercase",
-            "asciifolding",
+            "lowercasepreserveoriginal",
```
kpsherva, line 17: @wgresshoff could I ask you to move all the changes to funder-v3.0.0.json files?
in case of breaking changes like these we need to upgrade the version of the mapping, otherwise instance upgrades are complex

slint: Since in practice this doesn't break existing indices created with the old mapping but only affects search behavior/accuracy, I feel we can merge without bumping the mapping version.

This means that for deployment, one needs to either:

- Do the usual recreation of indices (with downtime)
- Do a [live migration](https://github.com/inveniosoftware/docs-invenio-rdm/issues/696) with some special handling since the old and new index name will be the same (only the timestamp suffix changes).

wgresshoff: We used the recreation of indices with downtime and the missing funders were reindexed without problem.

PR #421, opened 2024-10-14
jobs: apply code upgrades

ntarocco:


invenio_vocabularies/jobs.py
```diff
@@ -98,12 +70,9 @@ class ProcessRORFundersJob(ProcessDataStreamJob):
...
 
     @classmethod
-    def default_args(cls, job_obj, since=None, **kwargs):
+    def build_task_arguments(cls, job_obj, since=None, **kwargs):
         """Generate default job arguments here."""
-        if since is None and job_obj.last_runs["success"]:
-            since = job_obj.last_runs["success"].started_at
-        else:
-            since = since or datetime.datetime.now()
+        since = since or datetime.datetime.now()
```
kpsherva, line 75: we said to change it to None, if I remember correctly, no?
```suggestion
        since = since or None
```

ptamarit:
- [ ] Check the 2 tasks (compatible with jobs v1.0.0) added in #423

PR #423, opened 2024-10-14
jobs: add import awards OpenAIRE; Update CORDIS

yashlamba:


ptamarit (approved):

ptamarit:
Tested both jobs locally, works fine for me.

PR #425, opened 2024-10-15
fix: revert generic writer and define OpenAIRE awards writer logic & fix: exclude unknown fields when updating awards with subjects & 📦 release: v6.4.1

### LINKED ISSUE #417: Records searchable by specific funding programmes
Currently, records can be search by framework programmes:
* HE (Horizon Europe)
* H2020 (Horizon 2020)
* FP7

We should be able to also search records by specific programmes:
* Euratom Research and Training Programme (EURATOM)
* European Research Council (ERC)
* Marie Skłodowska-Curie Actions (MSCA)

ptamarit:
Partially fixes #417

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/datastreams.py
+++ b/invenio_vocabularies/contrib/affiliations/datastreams.py
@@ -9,8 +9,11 @@
 
 """Affiliations datastreams, transformers, writers and readers."""
 
+from copy import deepcopy
+
 from flask import current_app
 
+from ...datastreams import StreamEntry
 from ...datastreams.errors import TransformerError, WriterError
 from ...datastreams.transformers import BaseTransformer
 from ...datastreams.writers import ServiceWriter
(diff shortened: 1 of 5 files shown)


invenio_vocabularies/contrib/affiliations/datastreams.py
```diff
@@ -96,15 +99,24 @@ def _entry_id(self, entry):
...
 
-    def write(self, stream_entry, *args, **kwargs):
-        """Writes the input entry using a given service."""
-        entry = stream_entry.entry
-
-        return super().write(stream_entry, *args, **kwargs)
-
-    def write_many(self, stream_entries, *args, **kwargs):
-        """Writes the input entries using a given service."""
-        return super().write_many(stream_entries, *args, **kwargs)
```
ptamarit, line 107: `write` and `write_many` were not needed since the validation logic was moved to the transformer and these methods were simply calling their parent class.

invenio_vocabularies/datastreams/writers.py
```diff
@@ -87,23 +87,8 @@ def _resolve(self, id_):
...
-                else:
-                    combined_dict[key] = value
-            else:
-                combined_dict[key] = value
-
-        return StreamEntry(
-            self._service.update(self._identity, vocab_id, combined_dict)
-        )
+        updated = dict(current.to_dict(), **entry)
+        return StreamEntry(self._service.update(self._identity, vocab_id, updated))
```
ptamarit, line 91: Rolling back to the initial implementation (which overwrites keys at the top level), which was working for all cases, except for the OpenAIRE Affiliations, were we want to update/add identifiers to a list.

invenio_vocabularies/contrib/affiliations/datastreams.py
```diff
@@ -96,15 +99,24 @@ def _entry_id(self, entry):
...
+            # For each new identifier
+            for new_identifier in entry["identifiers"]:
+                # Either find an existing identifier with the same scheme and update the "identifier" value
+                for existing_identifier in updated["identifiers"]:
+                    if existing_identifier["scheme"] == new_identifier["scheme"]:
+                        existing_identifier["identifier"] = new_identifier["identifier"]
+                        break
+                # Or add the new identifier to the list of identifiers
+                else:
+                    updated["identifiers"].append(new_identifier)
```
ptamarit, line 117: We now have an implementation of the update logic very specific to the use case of updating/adding identifiers to a list of identifiers, making sure that we do not overwrite existing identifiers, and making sure that we do not end up with two identifiers for a given scheme.

slint (approved):

PR #431, opened 2024-11-15
subjects: fix search for MeSH

### LINKED ISSUE #430: bug: search doesn't work with MeSH
## Bug
After importing MeSH on local, it was noted that the search is not working for MeSH entries.

Data source: https://github.com/galterlibrary/invenio-subjects-mesh/tree/master/invenio_subjects_mesh/vocabularies

## Cause
The data doesn't have title which is required for subjects search. There can be 2 fixes:
- [ ] Add title from subject (MeSH transformer)
- [ ] Add subject field to search options

yashlamba:
Closes #430

### DIFF
--- a/invenio_vocabularies/contrib/subjects/config.py
+++ b/invenio_vocabularies/contrib/subjects/config.py
@@ -11,6 +11,7 @@
 """Subjects configuration."""
 
 from flask import current_app
+from invenio_i18n import get_locale
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services import SearchOptions
 from invenio_records_resources.services.records.components import DataComponent
@@ -22,6 +23,7 @@ from ...services.querystr import FilteredSuggestQueryParser
 subject_schemes = LocalProxy(
     lambda: current_app.config["VOCABULARIES_SUBJECTS_SCHEMES"]
 )
+localized_title = LocalProxy(lambda: f"title.{get_locale()}^20")
 
 
 class SubjectsSearchOptions(SearchOptions):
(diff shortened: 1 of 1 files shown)


invenio_vocabularies/contrib/subjects/config.py
```diff
@@ -31,6 +31,7 @@ class SubjectsSearchOptions(SearchOptions):
         filter_field="scheme",
         fields=[  # suggest fields
             "title.*^100",
```
slint, line 33: minor: actually looking at this, it's going to probably suffer from the same issue with the "too many search clauses (max 1024)" when searching for more than 3-4 words.

We should apply the same style as in [e.g. affiliations](https://github.com/inveniosoftware/invenio-vocabularies/blob/f36aa0abf273e46c0e2255a1481e89ba21bd46fe/invenio_vocabularies/contrib/affiliations/config.py#L24-L41), so the search order should be:

1. Generic/common title field (in this case `subject`)
2. Acronyms (N/A for subjects)
3. Localized title (i.e. `title.{user_locale}`)
4. Synonyms/aliases (in this case `synonyms`)

invenio_vocabularies/contrib/subjects/mesh/datastreams.py
```diff
@@ -22,14 +22,18 @@ def apply(self, stream_entry, *args, **kwargs):
...
         # We just want to use the ID prefixed by "mesh:""
         try:
             mesh_id = entry_data["id"].split("/")[-1]
+            entry_data["id"] = "mesh:" + mesh_id
         except Exception:
             raise TransformerError("Not a valid MeSH ID.")
 
-        entry_data["id"] = "mesh:" + mesh_id
+        entry_data["title"] = title = entry_data.get("title", {})
+        if "en" not in title:
```
slint, line 34: nit: add a comment here, that the MeSH import file comes with the English title by default

slint (approved):

PR #437, opened 2024-11-20
awards: add fields start/end date and description

yashlamba:


invenio_vocabularies/contrib/awards/datastreams.py
```diff
@@ -265,6 +272,7 @@ def apply(self, stream_entry, **kwargs):
             award["program"] = unique_programme_related_legal_basis["code"]
 
         stream_entry.entry = award
+        breakpoint()
```
ptamarit, line 275:
```suggestion
```

ptamarit (approved):

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -61,6 +61,12 @@ class AwardSchema(BaseVocabularySchema, ModePIDFieldVocabularyMixin):
 
     organizations = fields.List(fields.Nested(AwardOrganizationRelationSchema))
 
+    description = fields.String()
```
0einstein0, line 64:
```suggestion
    description = SanitizedUnicode()
```

0einstein0 (commented):
LGTM. left two suggestions, also the test needs to be updated

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -61,6 +61,12 @@ class AwardSchema(BaseVocabularySchema, ModePIDFieldVocabularyMixin):
...
+
+    start_date = fields.String()
+
```
0einstein0, line 67: Do we need to add these in `AwardRelationSchema` as well?

invenio_vocabularies/contrib/awards/datastreams.py
yashlamba: oops

invenio_vocabularies/contrib/awards/datastreams.py
```diff
@@ -117,6 +117,13 @@ def apply(self, stream_entry, **kwargs):
         if acronym:
             award["acronym"] = acronym
 
+        if "startDate" in record:
+            award["start_date"] = record["startDate"]
+        if "endDate" in record:
+            award["end_date"] = record["endDate"]
+        if "summary" in record:
+            award["description"] = record["summary"]
```
slint, line 125: minor: should be `description` (and also added to the JSONSchema and mappings)

invenio_vocabularies/contrib/awards/schema.py
slint: I would say no for now, since we don't have a use-case where we want to query records by their awards' start/end dates...

invenio_vocabularies/contrib/awards/schema.py
```diff
@@ -61,6 +61,12 @@ class AwardSchema(BaseVocabularySchema, ModePIDFieldVocabularyMixin):
 
     organizations = fields.List(fields.Nested(AwardOrganizationRelationSchema))
 
+    description = SanitizedUnicode()
+
+    start_date = fields.String()
```
slint, line 66: minor: this might be better to be `ISODateString` from marshmallow utils so that we do some validation on the string before indexing to search (which would fail for invalid ISO Date strings)

slint (approved):

slint (approved):

PR #439, opened 2024-11-22
datastream: optimize memory usage on ORCID sync

jrcastro2:


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -67,42 +68,55 @@ def _process_lambda_file(self, fileobj):
...
+                try:
+                    result = future.result()
+                    if result:
+                        yield result
+                finally:
+                    # Explicitly release memory, as we don't need the future anymore.
+                    # This is mostly required because as long as we keep a reference to the future
+                    # (in the above futures dict), the garbage collector won't collect it
+                    # and it will keep the memory allocated.
+                    futures.pop(future, None)
```
slint, line 119: minor: isn't it a bit dangerous to modify the `futures` dictionary while iterating over it? The alternative way would be to reverse the key/value above (i.e. `orcid: executor.submit(...)` inside the dict comprehension), and then iterate over `list(future.keys())` or similar.

invenio_vocabularies/cli.py
```diff
@@ -29,6 +29,8 @@ def _process_vocab(config, num_samples=None):
         readers_config=config["readers"],
         transformers_config=config.get("transformers"),
         writers_config=config["writers"],
+        batch_size=config.get("batch_size", 1000),
```
slint, line 32:
```suggestion
        batch_size=config.get("batch_size", 100),
```

minor: just because below I see the batch size being 100 by default. Otherwise 1000 should be applied everywhere? Or it could be I've misunderstood, and the 100 is just for chunking the 1000-sized batch, in which case ignore this comment completely :)

jrcastro2: It's the second case you described, `batch_size` is to determine the size of batch that is processed from reading to writting, while the 100 is just to chunk only the read process in smaller btaches (with this it seems to use a bit less memory)

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -67,42 +68,55 @@ def _process_lambda_file(self, fileobj):
...
                 )
                 for orcid in orcids
-            ]
-            for future in as_completed(futures):
-                result = future.result()
-                if result is not None:
-                    yield result
+            }
+
+            for orcid, future in list(futures.items()):
```
kpsherva, line 109: is casting to list needed?

slint: Could also be just iterating of `list(futures.keys())`, since the reason is that we're modifying the `futures` dictionary as we iterate over the keys.

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -67,42 +68,55 @@ def _process_lambda_file(self, fileobj):
         if self.since:
             time_shift = self.since
         last_sync = arrow.now() - timedelta(**time_shift)
+        try:
+            file_content = fileobj.read().decode("utf-8")
 
-        file_content = fileobj.read().decode("utf-8")
-
-        csv_reader = csv.DictReader(file_content.splitlines())
+            csv_reader = csv.DictReader(file_content.splitlines())
```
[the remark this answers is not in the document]
jrcastro2: Good point! Thanks, I updated it

PR #441, opened 2024-11-25
names: link affiliations IDs from ORCiD parsing

### LINKED ISSUE #442: names: link affiliatons when parsing ORCiD entries

slint:
- Closes #442.
- **names: extract affiliation identifiers from employments**
- **names: support mapping ORCiD org IDs to linked affiliation IDs**

### DIFF
--- a/invenio_vocabularies/config.py
+++ b/invenio_vocabularies/config.py
@@ -213,3 +213,16 @@ VOCABULARIES_ORCID_SYNC_SINCE = {
     "days": 1,
 }
 """ORCID time shift to sync. Parameters accepted are the ones passed to 'datetime.timedelta'."""
+
+VOCABULARIES_ORCID_ORG_IDS_MAPPING_PATH = None
+"""Path to the CSV file for mapping ORCiD organization IDs to affiliation IDs.
+
+The path can be specified as either an absolute path or a relative path within the
+Flask app instance folder (i.e. ``current_app.instance_path``).
+
+The CSV file should have the following columns:
+
+- `org_scheme`: The ORCiD organization ID.
+- `org_id`: The ORCiD organization ID.
+- `aff_id`: The affiliation ID to map to.
+"""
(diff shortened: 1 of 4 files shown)


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -48,7 +49,7 @@ def _fetch_orcid_data(self, orcid_to_sync, bucket):
         key = f"{suffix}/{orcid_to_sync}.xml"
         try:
             return self.s3_client.read_file(f"s3://{bucket}/{key}")
-        except Exception as e:
+        except Exception:
             # TODO: log
             return None
```
jrcastro2, line 54: Nothing to do with this PR, I think this is my leftover I will create a PR to fix this it shouldn't silently fail

jrcastro2 (approved):
Looks good to me! Thanks a lot! 🚀

invenio_vocabularies/config.py
```diff
@@ -213,3 +213,13 @@ def is_edmo(val):
...
+
+VOCABULARIES_ORCID_ORG_IDS_MAPPING_PATH = None
+"""Path to the CSV file for mapping ORCiD organization IDs to affiliation IDs.
+
+The CSV file should have the following columns:
+
+- `org_scheme`: The ORCiD organization ID.
+- `org_id`: The ORCiD organization ID.
+- `aff_id`: The affiliation ID to map to.
+"""
```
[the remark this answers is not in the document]
slint: Thanks, yeah the relative path was a last-minute addition :)

invenio_vocabularies/contrib/names/datastreams.py
slint: I've added `current_app.logger.exception(...)`, so we don't have a TODO for now. But yeah, if you know in what ways this can fail, it might be good to handle in a special way.

PR #443, opened 2024-11-25
Improve search accuracy for names, affiliations, funders

slint:
- Closes https://github.com/inveniosoftware/invenio-vocabularies/issues/438
- Depends on https://github.com/inveniosoftware/invenio-records-resources/pull/602
- **names: add affiliation acronym in mappings and schema**
  * Dereferences the affiliation `acronym` when indexing names and serving
    REST API results. This is useful for disambiguating authors in search.
- **contrib: improve search accuracy for names, funders, affiliations**

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/config.py
+++ b/invenio_vocabularies/contrib/affiliations/config.py
@@ -13,7 +13,9 @@ from invenio_i18n import get_locale
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services import SearchOptions
 from invenio_records_resources.services.records.components import DataComponent
-from invenio_records_resources.services.records.params import SuggestQueryParser
+from invenio_records_resources.services.records.queryparser import (
+    CompositeSuggestQueryParser,
+)
 from werkzeug.local import LocalProxy
 
 from ...services.components import PIDComponent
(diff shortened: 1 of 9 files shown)


invenio_vocabularies/contrib/affiliations/config.py
```diff
@@ -13,31 +13,36 @@
...
 
-    suggest_parser_cls = SuggestQueryParser.factory(
+    suggest_parser_cls = CompositeSuggestQueryParser.factory(
         fields=[
-            "name^100",
-            "acronym.keyword^100",
-            "acronym^40",
+            # We boost the acronym fields, since they're smaller words and are more
+            # likely to be used in a query.
+            "acronym.keyword^12",
```
jrcastro2, line 36: Question: Not sure I understand why we are boosting the fields here if in the end, we are removing it in the `parse` of `CompositeSuggestQueryParser`?

invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
```diff
@@ -125,6 +125,17 @@
             "type": "text",
             "analyzer": "accent_edge_analyzer",
             "search_analyzer": "accent_analyzer"
+          },
+          "acronym": {
+            "type": "text",
+            "analyzer": "accent_edge_analyzer",
```
jrcastro2, line 131: Comment: Checking the mappings seems that `identifiers` which are included in the fields to be searched, do not have the same analyzer as acronym and name, therefore I don't think that searching for Acronym and Identifier name will work. However, this might not be an important use case to cover. Up to you.

jrcastro2 (approved):
LGTU! Peer reviewed with @0einstein0 

Just left a couple of comments

invenio_vocabularies/contrib/affiliations/config.py
slint: We're not removing from all the `multi_match` queries, only the `cross_fields` type, since [it's recommended in the docs not to apply any boosting](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#cross-field-analysis:~:text=Note%20that%20cross_fields%20is%20usually%20only%20useful%20on%20short%20string%20fields%20that%20all%20have%20a%20boost%20of%201.%20Otherwise%20boosts%2C%20term%20freqs%20and%20length%20normalization%20contribute%20to%20the%20score%20in%20such%20a%20way%20that%20the%20blending%20of%20term%20statistics%20is%20not%20meaningful%20anymore.).

invenio_vocabularies/contrib/names/mappings/os-v1/names/name-v2.0.0.json
slint: `identifiers.identifier` is there so that one can search explicitly e.g. by ORCID (or other name identifiers), basically copy-paste what they're searching for in the searchbar

PR #455, opened 2024-12-19
names: remove internal_id

jrcastro2:


jrcastro2:
Not needed, we keep the `internal_id`.

Issue #459, opened 2025-02-03
Error when searching affiliations: too_many_nested_clauses


ntarocco:
TransportError(500, 'search_phase_execution_exception', 'too_many_nested_clauses: Query contains too many nested clauses; maxClauseCount is set to 1024')

Request: '/api/affiliations?size=20&suggest=The+Barcelona+Institute+of+Science+and+Technology'

Search params:
```
{page: 1, size: 20, sort: 'bestmatch', suggest: 'The Barcelona Institute of Science and Technology'}
```

Partial query:
```
{"query":{"bool":{"filter":[{"bool":{"must_not":[{"terms":{"tags":["unlisted"]}}]}}],"should":[{"multi_match":{"query":"The Barcelona Institute of Science and Technology","fields":["acronym.keyword","acronym","name","aliases","title.en","id","identifiers.identifier","country","country_name","types"],"operator":"and","type":"cross_fields","boost":3}},{"multi_match":{"query":"The Barcelona Institute of Science and Technology","fields":["acronym.keyword^50","acronym^10","name^10","aliases^5","title.en^2"
```

```
{error: {failed_shards: [{"index":"'cds-rdm-prod-affiliations-affiliation-v2.0.0-1728558302'","node":"'hGnt4dq-T66FzzECMrjZNg'","reason":"{'type': 'too_many_nested_clauses', 'reason': 'too_many_nested_clauses: Query contains too many nested clauses; maxClauseCount is set to 1024'}","shard":"0"}], grouped: True, phase: 'query', reason: 'all shards failed', root_cause: [{"reason":"'too_many_nested_clauses: Query contains too many nested clauses; maxClauseCount is set to 1024'","type":"'too_many_nested_clauses'"}], type: 'search_phase_execution_exception'}}
```

[this Sentry excp](https://cds-sentry.web.cern.ch/sentry/cds-rdm/issues/37846/?environment=production)

sakshamarora1:
This is happening because we are using wildcard search on the title subfields. Also, the [recent changes](https://github.com/inveniosoftware/invenio-records-resources/commit/96ff9ea10d4a4195d267d486ab2296e742fa2a41) increase the number of clauses. Currently, we have 126 languages in cds-rdm which means each new word in the search query increases the clauses because of the n-gram analyzer.

Also, bit weird but I did some trial and error on what works and what doesn't:
```
The In of Sci and te
```
[...]

PR #460, opened 2025-02-12
contrib: Remove redundant fields from affiliations, funders search

sakshamarora1:
closes: https://github.com/inveniosoftware/invenio-vocabularies/issues/459

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/config.py
+++ b/invenio_vocabularies/contrib/affiliations/config.py
@@ -42,12 +42,9 @@ class AffiliationsSearchOptions(SearchOptions):
             # Aliases can sometimes be shorter, so we boost them a bit.
             "aliases^5",
             localized_title,
-            "id^2",
             # Allow to search identifiers directly (e.g. ROR)
             "identifiers.identifier",
-            "country",
             "country_name",
-            "types",
         ],
     )
 
--- a/tests/contrib/affiliations/test_affiliations_resource.py
+++ b/tests/contrib/affiliations/test_affiliations_resource.py
@@ -114,6 +114,14 @@ def _create_affiliations(service, identity):
                 "en": "Northwestern University",
             },
         },
+        {
+            "acronym": "LONG",
+            "id": "LONG",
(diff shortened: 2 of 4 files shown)


invenio_vocabularies/contrib/affiliations/config.py
```diff
@@ -42,12 +42,9 @@ class AffiliationsSearchOptions(SearchOptions):
             # Aliases can sometimes be shorter, so we boost them a bit.
             "aliases^5",
             localized_title,
-            "id^2",
```
tmorrell, line 45: I think this is going to break searching by ROR. The test data might be misleading here…in real data the id isn’t the same as the acronym.

invenio_vocabularies/contrib/affiliations/config.py
```diff
@@ -42,12 +42,9 @@ class AffiliationsSearchOptions(SearchOptions):
...
             # Allow to search identifiers directly (e.g. ROR)
             "identifiers.identifier",
-            "country",
```
tmorrell, line 48: This might make it difficult to disambiguate funders with the same name but in different countries.

tests/contrib/affiliations/test_affiliations_resource.py
```diff
@@ -159,3 +167,12 @@ def test_affiliations_suggest_sort(
     assert res.json["hits"]["total"] == 2
     assert res.json["hits"]["hits"][0]["id"] == "nu"
     assert res.json["hits"]["hits"][1]["id"] == "cern"  # due to nucleaire
```
zubeydecivelek, line 169: Is it still possible to search with country? Could we add a test to make sure it's working?

invenio_vocabularies/contrib/affiliations/config.py
sakshamarora1: Thanks! I've added a test, it works

invenio_vocabularies/contrib/affiliations/config.py
sakshamarora1: So, the country is the abbreviation and the country_name is the full name of the country, which is what is shown in the UI too. I've also added a test for search with country.

tmorrell (approved):
Thanks for the tests. Looks good!

PR #468, opened 2025-03-26
logging: add basic logging for ORCID

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-jobs/issues/67

### DIFF
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -12,6 +12,7 @@ import csv
 import io
 import tarfile
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from contextvars import copy_context
 from datetime import timedelta
 from itertools import islice
 from pathlib import Path
@@ -43,17 +44,18 @@ class OrcidDataSyncReader(BaseReader):
         self.s3_client = S3OrcidClient()
         self.since = since
 
-    def _fetch_orcid_data(self, orcid_to_sync, bucket):
+    def _fetch_orcid_data(self, app, orcid_to_sync, bucket):
         """Fetches a single ORCiD record from S3."""
         # The ORCiD file key is located in a folder which name corresponds to the last three digits of the ORCiD
         suffix = orcid_to_sync[-3:]
         key = f"{suffix}/{orcid_to_sync}.xml"
+        app.logger.debug(f"Fetching ORCiD record: {key} from bucket: {bucket}")
         try:
             # Potential improvement: use the a XML jax parser to avoid loading the whole file in memory
             # and choose the sections we need to read (probably the summary)
             return self.s3_client.read_file(f"s3://{bucket}/{key}")
         except Exception:
-            current_app.logger.exception("Failed to fetch ORCiD record.")
+            app.logger.exception(f"Failed to fetch ORCiD record: {key}")
 
     def _process_lambda_file(self, fileobj):
         """Process the ORCiD lambda file and returns a list of ORCiDs to sync.
@@ -87,7 +89,11 @@ class OrcidDataSyncReader(BaseReader):
                     )
 
                 if last_modified_date < last_sync:
+                    current_app.logger.debug(
+                        f"Skipping ORCiD {orcid} (last modified: {last_modified_date})"
+                    )
                     break
+                current_app.logger.debug(f"Yielding ORCiD {orcid} for sync.")
                 yield orcid
         finally:
             fileobj.close()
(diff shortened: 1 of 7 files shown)


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -43,17 +44,18 @@ def __init__(self, origin=None, mode="r", since=None, *args, **kwargs):
...
         suffix = orcid_to_sync[-3:]
         key = f"{suffix}/{orcid_to_sync}.xml"
+        flask_app.logger.debug(f"Fetching ORCiD record: {key} from bucket: {bucket}")
         try:
             # Potential improvement: use the a XML jax parser to avoid loading the whole file in memory
             # and choose the sections we need to read (probably the summary)
             return self.s3_client.read_file(f"s3://{bucket}/{key}")
         except Exception:
-            current_app.logger.exception("Failed to fetch ORCiD record.")
+            flask_app.logger.exception(f"Failed to fetch ORCiD record: {key}")
```
kpsherva, line 58: in your other [PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1989) you are using current_app, could you explain why we need flask app here?

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -97,10 +103,13 @@ def _iter(self, orcids):
         with ThreadPoolExecutor(
             max_workers=current_app.config["VOCABULARIES_ORCID_SYNC_MAX_WORKERS"]
         ) as executor:
+            app = current_app._get_current_object()
```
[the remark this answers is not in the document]
jrcastro2: We want to do this to pass it to propagate it to the threads as they are not aware otherwise

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: This is being executed in a thread outside the app context, we need to pass it as an argument, I called it flask_app to avoid confusion but is the same as current_app

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -97,10 +103,13 @@ def _iter(self, orcids):
...
             # futures is a dictionary where the key is the ORCID value and the item is the Future object
             futures = {
                 orcid: executor.submit(
+                    copy_context().run,  # Required to pass the context to the thread
```
jrcastro2, line 110: add some docstrings to explain it

invenio_vocabularies/contrib/names/datastreams.py
utnapischtim: in my thinking, normally invenio calls such variables only `app`. naming it `flask_app` could add more confusion than resolving confusion

invenio_vocabularies/contrib/names/datastreams.py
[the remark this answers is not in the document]
utnapischtim: what is the difference between `current_app`, `current_app._get_current_object()` why was it not enough to only propagate `current_app` to the thread? is this because of the app context?  so that `current_app` also given by parameter would not work, because of the missing app_context?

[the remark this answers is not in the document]
jrcastro2: `current_app` is a proxy that only works within an active Flask app context. `_get_current_object()` returns the actual Flask app instance behind the proxy. In threads, the context isn’t propagated, so the proxy breaks — that’s why we pass the real object using `_get_current_object()`, which is safe to use in threads.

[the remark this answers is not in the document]
utnapischtim: thanks for clarification

invenio_vocabularies/contrib/names/datastreams.py
jrcastro2: My idea was to differentiate it since it’s not the proxy we typically use, but the actual Flask app instance passed to the thread. That said, I’m happy to rename it to `app` if you think it’s clearer and more consistent.

utnapischtim: if we compare it with what we have on the `ext.py` files like [here](https://github.com/inveniosoftware/invenio-jobs/blob/master/invenio_jobs/ext.py#L40). the `app` there and the `flask_app` are the type wise the same?

jrcastro2: Right, they are the same. In `ext.py`, it’s clear we’re working with the Flask app since it’s being instantiated there. Here, it’s a bit less obvious that we’re dealing with a Flask-based app. In any case, I’ve updated the name for consistency.

PR #472, opened 2025-04-17
readers: allow setting user_agent in SPARQLReader

SarahW91:
### Description

The invenio-vocabularies SPARQLReader so far doesn't allow setting a user-agent string.

This is however required to use e.g. wikidata's SPARQL endpoint as per their [policy](https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_User-Agent_Policy).

After this PR is merged, it will allow setting a user agent optionally when configuring the SPARQLReader.

### Checklist

### DIFF
--- a/invenio_vocabularies/datastreams/readers.py
+++ b/invenio_vocabularies/datastreams/readers.py
@@ -407,13 +407,13 @@ class RDFReader(BaseReader):
 class SPARQLReader(BaseReader):
     """Generic reader class to fetch and process RDF data from a SPARQL endpoint."""
 
-    def __init__(self, origin, query, client_params=None, mode="r", *args, **kwargs):
+    def __init__(self, origin, query, mode="r", client_params=None, *args, **kwargs):
         """Initialize the reader with the data source.
 
         :param origin: The SPARQL endpoint from which to fetch the RDF data.
         :param query: The SPARQL query to execute.
-        :client_params: Additional client parameters to pass to the SPARQL client.
         :param mode: Mode of operation (default is 'r' for reading).
+        :param client_params: Additional client parameters to pass to the SPARQL client.
         """
         self._origin = origin
         self._query = query
--- a/invenio_vocabularies/datastreams/readers.py
+++ b/invenio_vocabularies/datastreams/readers.py
@@ -281,15 +281,15 @@ class OAIPMHReader(BaseReader):
     """OAIPMH reader."""
 
     def __init__(
-        self,
-        *args,
-        base_url=None,
-        metadata_prefix=None,
-        set=None,
-        from_date=None,
-        until_date=None,
-        verb=None,
-        **kwargs,
+            self,
+            *args,
+            base_url=None,
+            metadata_prefix=None,
+            set=None,
+            from_date=None,
+            until_date=None,
+            verb=None,
+            **kwargs,
     ):
         """Constructor."""
         self._base_url = base_url
@@ -384,7 +384,7 @@ class RDFReader(BaseReader):
     def _iter(self, rdf_graph):
         """Iterate over the RDF graph, yielding one subject at a time."""
         for subject, _, _ in rdf_graph.triples(
-            (None, rdflib.RDF.type, self.skos_core.Concept)
+                (None, rdflib.RDF.type, self.skos_core.Concept)
         ):
             yield {"subject": subject, "rdf_graph": rdf_graph}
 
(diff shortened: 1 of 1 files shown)


tmorrell (approved):
Looks like a good improvement. This will need one more review, ideally from someone using the sparql_client to make sure "" as the agent doesn't break existing use cases.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -416,6 +416,8 @@ def __init__(self, origin, query, mode="r", *args, **kwargs):
         """
         self._origin = origin
         self._query = query
+        self._user_agent = kwargs.get("user_agent", "")
```
slint, line 419:
```suggestion
        self._client_params = client_params or {}
```

minor: I would convert this to a `client_params` dictionary-style value (and add it as a param in `SPARQLReader.__init__`), so that we can pass arbitrary arguments in the future. This is also to allow just not passing `user_agent` at all, so that the default value from the `SPARQLWrapper` class.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -430,7 +432,7 @@ def read(self, item=None, *args, **kwargs):
                 "SPARQLReader does not support being chained after another reader"
             )
 
-        sparql_client = sparql.SPARQLWrapper(self._origin)
+        sparql_client = sparql.SPARQLWrapper(self._origin, agent=self._user_agent)
```
slint, line 435:
```suggestion
        sparql_client = sparql.SPARQLWrapper(self._origin, agent=**self._client_params)
```

(see above comment)

slint (approved):
LGTM, one minor improvement to possibly allow extending more in the future.

invenio_vocabularies/datastreams/readers.py
SarahW91: It doesn't quite work with the code you provided. Only alternative I found when using `client_params` would be to conditionally set the agent when initializing the `SPARQLWrapper` to avoid overriding the default value:
```
  if self._client_params.get("user_agent"):
      sparql_client = sparql.SPARQLWrapper(self._origin, agent=self._client_params.get("user_agent"))
  else:
      sparql_client = sparql.SPARQLWrapper(self._origin)
```

invenio_vocabularies/datastreams/readers.py
SarahW91: In principle I like this idea! See comment below for my issues with implementation

invenio_vocabularies/datastreams/readers.py
tmorrell: I think that's safer, since then it wouldn't impact the default. New parameters would still need to be added to this "if" block, but that seems fine.

SarahW91: I implemented it this way now. Merge whenever you feel like it's appropriate ;)

invenio_vocabularies/datastreams/readers.py
```diff
@@ -407,15 +407,18 @@ def read(self, item=None, *args, **kwargs):
 class SPARQLReader(BaseReader):
     """Generic reader class to fetch and process RDF data from a SPARQL endpoint."""
 
-    def __init__(self, origin, query, mode="r", *args, **kwargs):
+    def __init__(self, origin, query, client_params=None, mode="r", *args, **kwargs):
```
tmorrell, line 410: I think you want to put `client_params` after `mode`. That way if someone hasn't passed the mode label it won't break.

invenio_vocabularies/datastreams/readers.py
```diff
@@ -407,15 +407,18 @@ def read(self, item=None, *args, **kwargs):
...
         """Initialize the reader with the data source.
 
         :param origin: The SPARQL endpoint from which to fetch the RDF data.
         :param query: The SPARQL query to execute.
+        :client_params: Additional client parameters to pass to the SPARQL client.
```
tmorrell, line 415: Looks like you're missing a `param` here. If you decide to change the order in the function definition, also change the order here.

SarahW91: Good catch, thanks!

invenio_vocabularies/datastreams/readers.py
SarahW91: I followed that suggestion in my latest commit

PR #487, opened 2025-05-21
make it possible to search over id in affiliations, awards and funders

Ducica:

### Description

> Please describe briefly your pull request.

We have received  feedback from our users, that they would really appreciate, if it would be possible to search via id in some specialized vocabularies.

For example, ids of funders in Czech Republic are very well known and in many cases it would be easier for users to look for it based on id, than based on name or title or something like that. 

I don't think this PR bears any negative impact on the library and for us it would be useful to have it. Please if you can take a look and let us know if you would merge this. Thanks!

**Reminder**

### DIFF
--- a/invenio_vocabularies/contrib/affiliations/config.py
+++ b/invenio_vocabularies/contrib/affiliations/config.py
@@ -45,6 +45,7 @@ class AffiliationsSearchOptions(SearchOptions):
             # Allow to search identifiers directly (e.g. ROR)
             "identifiers.identifier",
             "country_name",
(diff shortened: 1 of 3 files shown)


Ducica:
@tmorrell Thanks Tom for a quick reply. Yes, we have simply vocabularies where ids are actually not in identifiers. And in some cases, it might not make fully sense to have them in identifiers. However, maybe I was hasty with opening this PR. Let me make some more consultations please and then I will maybe close this PR during next week.

PR #494, opened 2025-07-01
names: orcid job: fix since argument not working

sakshamarora1:


invenio_vocabularies/jobs.py
```diff
@@ -173,4 +171,9 @@ class ImportORCIDJob(ProcessDataStreamJob):
     @classmethod
     def build_task_arguments(cls, job_obj, since=None, **kwargs):
         """Process ORCID data."""
-        return {"config": {**ORCID_PRESET_DATASTREAM_CONFIG}}
+        task_args = {"config": {**ORCID_PRESET_DATASTREAM_CONFIG}}
+        for reader in task_args["config"]["readers"]:
+            if reader["type"] == "orcid-data-sync":
```
kpsherva, line 176: @ntarocco let's cross check it with your PR, maybe we could finalise the change together with what @sakshamarora1 proposes here

invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -69,9 +69,9 @@ def _process_lambda_file(self, fileobj):
         date_format = "YYYY-MM-DD HH:mm:ss.SSSSSS"
         date_format_no_millis = "YYYY-MM-DD HH:mm:ss"
         time_shift = current_app.config["VOCABULARIES_ORCID_SYNC_SINCE"]
-        if self.since:
-            time_shift = self.since
         last_sync = arrow.now() - timedelta(**time_shift)
+        if self.since:
+            last_sync = arrow.get(self.since)
```
palkerecsenyi, line 74: @sakshamarora1 I'm working on removing the string casts in https://github.com/inveniosoftware/invenio-jobs/issues/87, maybe we won't need the `arrow.get` call as it's already a Python datetime? Not sure how arrow works in general so I could be wrong.

invenio_vocabularies/jobs.py
ntarocco: @sakshamarora1 @kpsherva @palkerecsenyi do we merge this one or not?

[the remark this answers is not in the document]
sakshamarora1: Because, we're using the `ORCID_PRESET_DATASTREAM_CONFIG` only in this reader, for all the other ones, the `since` argument is already present. I'm not sure as to why it was done in this way in the first place, probably to have control via the config variable.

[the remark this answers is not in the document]
kpsherva: > Because, we're using the `ORCID_PRESET_DATASTREAM_CONFIG` only in this reader, for all the other ones, the `since` argument is already present. I'm not sure as to why it was done in this way in the first place, probably to have control via the config variable.

test

kpsherva (approved):

Issue #496, opened 2025-07-03
Filtering on scheme in subjects search doesn't work


sakshamarora1:
Package version (if known): 8.0.1

## Describe the bug
Filtering via UI dropdown is not working. The `FilteredSuggestQueryParser` was deprecated, but [on the UI, the query formation wasn't updated to follow the changes in backend. ](https://github.com/inveniosoftware/react-invenio-forms/blob/8b7e578c948cc6fc05a4f0431a955c49f130aaa3/src/lib/forms/widgets/custom_fields/SubjectAutocompleteDropdown.js#L20)

## Steps to Reproduce
1. Have at least 2 schemes registered via `VocabularyScheme` in your instance
2. Go to deposit form and select a scheme
3. See error, search is broken

## Expected behavior

2 ways forward:
1. Scheme filtering should be added back something similar to the depecated `FilteredSuggestQueryParser`
OR
2. We should remove scheme selection and update scheme field to have a text field with analyzers for autocomplete to work

PR #497, opened 2025-07-04
tests: subjects: Add filtered term search checks

sakshamarora1:
fixes: #496


sakshamarora1:
The tests are dependent on https://github.com/inveniosoftware/invenio-records-resources/pull/640

PR #498, opened 2025-07-07
fix(jobs): remove string casts from `since`

palkerecsenyi:
**Please release this PR at the same time as #494**

**Please release this PR at the same time as the changes in https://github.com/inveniosoftware/invenio-jobs/issues/87#issuecomment-3045453052**

Closes https://github.com/inveniosoftware/invenio-jobs/issues/87

---

* Jobs currently use a string-cast `since` parameter rather than keeping it as the Python `datetime` obj due to a previous issue that wasn't fully investigated

* Provisionally removed the string casts which need to be tested. This PR is a draft until I've tested compatibility. Some changes in CDS will be needed.

### DIFF
--- a/invenio_vocabularies/contrib/common/ror/datastreams.py
+++ b/invenio_vocabularies/contrib/common/ror/datastreams.py
@@ -55,7 +55,6 @@ class RORHTTPReader(BaseReader):
                     last_dump_date = json_ld_data.get(
                         "dateCreated"
                     ) or json_ld_data.get("datePublished")
-                    print(last_dump_date)
                     last_dump_date = datetime.fromisoformat(
                         last_dump_date.replace("Z", "+00:00")
                     )


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -82,9 +83,11 @@ def _process_lambda_file(self, fileobj):
...
                 last_modified_str = row["last_modified"]
                 try:
-                    last_modified_date = arrow.get(last_modified_str, date_format)
-                except arrow.parser.ParserError:
-                    last_modified_date = arrow.get(
+                    last_modified_date = datetime.strptime(
+                        last_modified_str, date_format
+                    )
+                except Exception:
+                    last_modified_date = datetime.strptime(
```
kpsherva, line 90: what is the reason for not using arrow package?

palkerecsenyi: We were already using arrow for only a very small number of cases that could be nearly fully handled using only Python standard library methods from `datetime` so it seemed nicer to get rid of a dependency here

invenio_vocabularies/contrib/common/ror/datastreams.py
```diff
@@ -52,9 +52,12 @@ def _get_last_dump_date(self, linksets):
                     json_ld_reponse.raise_for_status()
                     json_ld_data = json_ld_reponse.json()
 
-                    last_dump_date = arrow.get(
-                        json_ld_data.get("dateCreated")
-                        or json_ld_data.get("datePublished")
+                    last_dump_date = json_ld_data.get(
+                        "dateCreated"
+                    ) or json_ld_data.get("datePublished")
+                    print(last_dump_date)
```
kpsherva, line 58: 👀

PR #529, opened 2026-03-03
fix(names): skip invalid ORCID entries

Samk13:

### Description

closes https://github.com/inveniosoftware/invenio-app-rdm/issues/3351

This fixes the leak of raw ORCID records when `person.name` or `family-name` is missing.

Previously, those records were only marked filtered, but the datastream skips transformed entries based on errors, so they could still be written out unchanged.
Now we append a TransformerError and return early, which makes the pipeline correctly skip them.

A regression test was added to ensure missing-name ORCID records always produce errors and are not treated as valid transformed output.

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021-2026 CERN.
+# Copyright (C) 2026 KTH Royal Institute of Technology.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
@@ -276,10 +277,11 @@ class OrcidTransformer(BaseTransformer):
         given_names = name.get("given-names", None) if name else None
 
         if name is None or family_name is None:
-            current_app.logger.warning(
-                f"Missing name or family name for ORCID ID: {orcid_id}"
+            error = TransformerError(
+                f"Missing name or family name for ORCiD ID: {orcid_id}."
             )
-            stream_entry.filtered = True
+            stream_entry.errors.append(error)
+            current_app.logger.warning(error)
             return stream_entry
 
         full_name = " ".join(
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 ..
     Copyright (C) 2020-2025 CERN.
     Copyright (C) 2024-2026 Graz University of Technology.
+    Copyright (C) 2026 KTH Royal Institute of Technology.
 
     Invenio-Vocabularies is free software; you can redistribute it and/or
     modify it under the terms of the MIT License; see LICENSE file for more
@@ -8,6 +9,10 @@
 
 Changes
 =======
+
+Version v11.1.1 (released 2026-04-23)
+
+- fix(names): skip invalid ORCID entries
 
 Version v11.1.0 (released 2026-04-22)
 
--- a/invenio_vocabularies/__init__.py
+++ b/invenio_vocabularies/__init__.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020-2026 CERN.
 # Copyright (C) 2024-2026 Graz University of Technology.
+# Copyright (C) 2026 KTH Royal Institute of Technology.
 #
 # Invenio-Vocabularies is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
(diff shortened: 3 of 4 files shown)


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -276,10 +277,11 @@ def apply(self, stream_entry, **kwargs):
         given_names = name.get("given-names", None) if name else None
 
         if name is None or family_name is None:
-            current_app.logger.warning(
-                f"Missing name or family name for ORCID ID: {orcid_id}"
+            error = TransformerError(
+                f"Missing name or family name for ORCiD ID: {orcid_id}."
             )
```
slint, line 282: minor: I'm a bit unhappy with how we so far construct errors and log messages, since we don't make use of `extra` for structured data, which is a much better approach for filtering or indexing application logs.

I would have preferred that we have a stable message and the specific `orcid_id` is an attribute of the class. This means though that we either make custom `TransformError` subclasses with the attributes we care about (e.g. in this case `ORCIDTransformError("...", orcid_id=orcid_id)`) or that we just accept a generic `**extra`/`**context` in `TransformError`.

Then when logging we would do something like:


```python
current_app.logger.warning(str(error), extra={"orcid_id": error.orcid_id})
# ...or...
current_app.logger.warning(str(error), extra=error.extra)
```

---

Some context on why this comment, is that we currently have in our Sentry instance 100k+ unique events for each vocabulary transform failure. Normally [Sentry groups/fingerprints](https://docs.sentry.io/concepts/data-management/event-grouping/), warning and errors, but in this case it cannot, since the message varies always...

slint: And obviously, this here is not the only offender of this "stable error message" principle, but I feel that if we don't start doing something about it (e.g. [adding relevant linting rules](https://pypi.org/project/flake8-logging-format/)), we'll just continue doing it.

Samk13: The initial scope of this PR was to make the converter produce usable data, so for error handling, I started from the patterns we already have.

With [this second commit change](https://github.com/inveniosoftware/invenio-vocabularies/pull/529/commits), I tried to address the specific concern you raised for this ORCiD path: keeping the error/log message stable and moving the orcid_id into a structured context so grouping works better. 

If this matches what you had in mind, I’d prefer to handle the broader cleanup and any linting/enforcement for the wider codebase in a separate follow-up PR after this one is merged. 
WDYT?

slint: To be clear, I'm not 100% sure which of the two patterns is better, and I should have probably tagged my comment with `minor/shelve: ...`. So https://github.com/inveniosoftware/invenio-vocabularies/pull/529/commits/3a6257f631157f001cc3b0e73039539794553f88 can be omitted in this PR until we make a wider decision.

I'll write up an issue with the options and share in the maintainers meeting next week.

Samk13: Omitted that last change and moved it into a [separate draft](https://github.com/inveniosoftware/invenio-vocabularies/pull/534) so we can discuss it further in the upcoming meeting.

Samk13:
I think this may have slipped through, and the issue still seems to be present (unless it’s been addressed in another issue). @slint, could you let me know if there’s anything you’d like adjusted here?

Otherwise, it would be good to move this forward one way or the other so it doesn’t remain open.

slint (approved):

slint:
> I think this may have slipped through...

Very much true 😅

Feel free to merge + release

PR #537, opened 2026-04-20
fix(datastreams): Log ORCID read errors as warnings

### LINKED ISSUE #511: Avoid Orcid spamming on controlled errors
On the ORCID importer we want to avoid spamming on sentry when there are controlled errors such as the [following ones](https://github.com/inveniosoftware/invenio-vocabularies/blob/feature/subtasks-tracking/invenio_vocabularies/contrib/names/datastreams.py#L278-L308). All these errors should be added to an `warnings` attribute that should be added to the [`Datastream` class](https://github.com/inveniosoftware/invenio-vocabularies/blob/feature/subtasks-tracking/invenio_vocabularies/datastreams/datastreams.py#L38). And instead of triggering errores, we should trigger the [warnings logging here.](https://github.com/inveniosoftware/invenio-vocabularies/blob/06c505aeba8f205919e2ac3902dc59a7bbc16d53/invenio_vocabularies/services/tasks.py#L31)

sakshamarora1:
Fixes: #511 

* This will fix the issue of read errors getting logged to sentry due to wrongly being marked as errors. Because we should only mark issues as ERROR/EXCEPTION when there are bugs in the code but most read errors are due to the data from ORCID being incomplete

### DIFF
--- a/invenio_vocabularies/contrib/names/datastreams.py
+++ b/invenio_vocabularies/contrib/names/datastreams.py
@@ -54,7 +54,7 @@ class OrcidDataSyncReader(BaseReader):
             # and choose the sections we need to read (probably the summary)
             return self.s3_client.read_file(f"s3://{bucket}/{key}")
         except Exception:
-            app.logger.exception(f"Failed to fetch ORCiD record: {key}")
+            app.logger.exception("Failed to fetch ORCiD record: %s", key)
 
     def _process_lambda_file(self, fileobj):
         """Process the ORCiD lambda file and returns a list of ORCiDs to sync.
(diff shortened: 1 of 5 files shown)


invenio_vocabularies/contrib/names/datastreams.py
```diff
@@ -134,7 +134,7 @@ def _iter(self, orcids):
                         )
                         yield result
                 except Exception:
-                    current_app.logger.exception(
+                    current_app.logger.warning(
```
slint, line 137: We gotta stop doing f-strings in logging since this is what causes Sentry to create individual issues out of each warning/error message.

We should either use the `extra` param or positional arg-based formatting.

slint (commented):
Fixing the Sentry spamming actually should be done by not using f-strings. I commented in one case, but it applies to all the new/changed logging calls in this PR

kpsherva (approved):

slint (approved):

PR #546, opened 2026-06-08
fix(datastreams): fix OAIPMHReader default parameters

egabancho:
* `metadata_prefix` and `verb` have a typo in their ternary expressions evaluating to `None` if no value is passed to the constructor, causing the defaults `"oai_dc"` and `"ListRecords"` to never be applied.

### DIFF
--- a/invenio_vocabularies/datastreams/readers.py
+++ b/invenio_vocabularies/datastreams/readers.py
@@ -293,11 +293,11 @@ class OAIPMHReader(BaseReader):
     ):
         """Constructor."""
         self._base_url = base_url
-        self._metadata_prefix = metadata_prefix if not None else "oai_dc"
+        self._metadata_prefix = metadata_prefix or "oai_dc"
         self._set = set
         self._until = until_date
         self._from = from_date
-        self._verb = verb if not None else "ListRecords"
+        self._verb = verb or "ListRecords"
         super().__init__(*args, **kwargs)
 
     def _iter(self, scythe, *args, **kwargs):


invenio_vocabularies/datastreams/readers.py
```diff
@@ -293,11 +293,13 @@ def __init__(
...
         self._base_url = base_url
-        self._metadata_prefix = metadata_prefix if not None else "oai_dc"
+        self._metadata_prefix = (
+            metadata_prefix if metadata_prefix is not None else "oai_dc"
+        )
         self._set = set
         self._until = until_date
         self._from = from_date
-        self._verb = verb if not None else "ListRecords"
+        self._verb = verb if verb is not None else "ListRecords"
```
utnapischtim: nit: could be written as `verb or "ListRecords"` and `metadata_prefix or "oai_dc"` to make it shorter. [ruff](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/)

utnapischtim (approved):

invenio_vocabularies/datastreams/readers.py
egabancho: In general, I don't like the `or` binary operator. It somewhat hides the fact that it treats all falsy values (`""`, `0`, `False`, `None`) the same, even though they are semantically different sometimes.

That said, in this particular case, there's no valid empty OAI-PMH verb or metadata prefix, so both approaches should be fine.

utnapischtim: i have also mixed feelings with the `or` operator in such cases. But for me the construct `verb if verb else "ListRecords"` is unnecessary verbose. `verb or "ListRecords"` reads nice: "this or the other".

so as you prefer :)

PR #563, opened 2026-08-04
fix(contrib): display all funders api search results

edivalentinitu:
* without the search prop, semantic-ui does extra filtering on the API hits, thus neglecting the already in place backend search filters
  * re-add the search prop with a twist - force recomputation of the options array because otherwise the options list will be polluted with "Add " options (again added by the semantic-ui component from allowAdditions)
  * the end result is displaying only one Add option + all the API search hits

Before:

https://github.com/user-attachments/assets/88d3413f-8a83-4c99-8a8a-4dc685c51323

After:

https://github.com/user-attachments/assets/c5bfc0c4-632f-4de5-8d93-acb81f431e8b


slint (approved):
LGTM! My only comment is about visual consistency with the affiliations search (which has the same fields since it's coming from ROR):

Image

Mainly this means that in the subtitle we should show the `{City}, {Country} – {Type}` instead of just the country code.

Regarding the display of the ROR ID, we should follow [the ROR display guidelines](https://ror.readme.io/docs/display), and I guess the [Short ROR ID](https://ror.readme.io/docs/display#short-ror-id).

PR #566, opened 2026-08-12
feat(awards): add website field populated from CORDIS

slint:
* Awards had nowhere to record a project's own website. The
  `identifiers` list cannot hold one, since only a single identifier per
  scheme is allowed and `url` should be used for the CORDIS project link.


mairasalazar (approved):
LGTM!

